PR #1, opened 2024-10-24
[RSPEED-187] Add precommit and github workflow

hosekadam:
Add pre-commit and github actions workflows.

Also set the target python version to 3.9 as it is default version
shipped with RHEL9 and our main dependencies requires older version.


.github/workflows/ruff.yml
```diff
@@ -0,0 +1,14 @@
+name: Ruff
+
+on:
+  push:
+    branches: [add-github-workflows]
+  pull_request:
+    branches: [add-github-workflows]
```
andywaltlova, line 7: Not sure if we won't need the `*` here to trigger it (push for main should be fine, but pull requests can come from any branch)

```suggestion
    branches: 
      - "*"
```

andywaltlova: Same for unit tests

hosekadam: Yeah, this was part of testing (but it didn't work :(). Originally, I had there `main` branch

hosekadam:
@andywaltlova Now it's ready for review - test PR with the action runs is here: https://github.com/hosekadam/digital-roadmap-backend/pull/2

When we merge that, all the actions should run whenever we create PR for main or push directly to main. The unit tests and codecov is failing on missing tests. As we create them, we might need to update the paths.

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,38 @@
...
+        os: [ubuntu-latest]
+    env:
+      OS: ${{ matrix.os }}
+      PYTHON: '3.10'
+    steps:
+    - uses: actions/checkout@main
+    - name: Setup Python
+      uses: actions/setup-python@main
+      with:
+        python-version: '3.10'
```
andywaltlova, line 22: I am fine with having codecov separate (even if that means the unit test run twice, once in this workflow and once in unit tests workflow), but we need to sync python version between them (will start thread on slack)

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,38 @@
...
+        pip install pytest
+        pip install pytest-cov
+        pytest --cov=./ --cov-report=xml
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v4
+      with:
+        directory: ./coverage/reports/
+        env_vars: OS,PYTHON
+        fail_ci_if_error: true
+        files: ./coverage1.xml,./coverage2.xml,!./cache
```
andywaltlova, line 34: I don't think we need special named files, or do we? I though the `--cov-report=xml` genrated in default file `coverage.xml`

andywaltlova: Screenshot from your workflow

image

andywaltlova (changes requested):
Looks good, I would just change the coverage file name and sync python versions, we can update the workflows when we have some tests.

.github/workflows/codecov.yml
hosekadam: I will remove this line about files

andywaltlova (approved):

PR #14, opened 2025-01-15
[RSPEED-427] Add systems endpoint

hosekadam:
Add endpoint for returning RHEL systems start/end dates

TODO:
- [x] create the unit tests more complex - just tried how them works
- [x] comment the code
- [x] improve the endpoints using some FastAPI features, I just basically test how this works

### DIFF
--- a/app/v1/lifecycle/systems/endpoints.py
+++ b/app/v1/lifecycle/systems/endpoints.py
@@ -1,0 +1,40 @@
+from fastapi import APIRouter, Path
+
+from app.data.systems import OS_DATA_MOCKED
+
+v1_router = APIRouter()
+
+
+@v1_router.get("/systems")
+async def get_systems():
+    systems = get_systems_data()
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+@v1_router.get("/systems/{major}")
+async def get_systems_major(major: int = Path(..., description="Major version number")):
+    systems = get_systems_data(major)
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+@v1_router.get("/systems/{major}/{minor}")
+async def get_systems_major_minor(
+    major: int = Path(..., description="Major version number"),
+    minor: int = Path(..., description="Minor version number"),
+):
+    systems = get_systems_data(major, minor)
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+def get_systems_data(major=None, minor=None):
+    data = OS_DATA_MOCKED
+
+    if major is not None:
+        data = [d for d in data if d["major"] == major]
+    if minor is not None:
+        data = [d for d in data if d["minor"] == minor]
+
+    return data
--- a/app/data/systems.py
+++ b/app/data/systems.py
@@ -1,0 +1,40 @@
+from datetime import date
+
+OS_DATA_MOCKED = [
+    {
+        "name": "RHEL",
+        "major": 9,
+        "minor": 2,
+        "release_date": date(2023, 5, 1),
+        "retirement_date": date(2023, 11, 1),
+        "systems": 5,
+        "lifecycle_type": "mainline",
+    },
+    {
+        "name": "RHEL",
+        "major": 8,
+        "minor": 3,
+        "release_date": date(2020, 11, 1),
+        "retirement_date": date(2021, 5, 1),
+        "systems": 50,
+        "lifecycle_type": "eus",
+    },
+    {
+        "name": "RHEL",
+        "major": 8,
+        "minor": 7,
+        "release_date": date(2023, 5, 1),
+        "retirement_date": date(2023, 5, 1),
+        "systems": 12,
+        "lifecycle_type": "e4s",
+    },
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -46,3 +46,7 @@ test:
 .PHONY: build
 build:
 	docker build -t digital_roadmap:latest -f Containerfile .
+
+.PHONY: tests
+tests:
+	$(VENV_DIR)/bin/pytest
```
samdoran, line 52: I think this does the same thing as the test target.

app/v1/lifecycle/systems/endpoints.py
```diff
@@ -0,0 +1,71 @@
+from datetime import date
+
+from fastapi import APIRouter
+
+v1_router = APIRouter()
+
+OS_DATA_MOCKED = [
```
samdoran, line 7: Put this in `data/[something].py` similar to [this](https://github.com/RedHatInsights/digital-roadmap-backend/pull/16/files#diff-28e5486514290d044ceb29f95bba42b2c4e2db5767b13777cff5522bc558f7c6). That helps to keep the data at least somewhat separate from the code.

samdoran (commented):
Looking good so far!

Makefile
hosekadam: That's right, I missed the target `test`. As I was used to using `tests` (which missed), I didn't look for other variants

Removed :)

.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
     rev: v0.7.0
     hooks:
       - id: ruff
+        args: [ --fix ]
```
samdoran, line 6: This can be done in a separate PR.

app/v1/lifecycle/systems/endpoints.py
```diff
@@ -0,0 +1,42 @@
+from fastapi import APIRouter, Path
+
+from app.data.systems import OS_DATA_MOCKED
+
+v1_router = APIRouter()
+
+
+@v1_router.get("/systems")
```
samdoran, line 8: nit: REST API endpoints should have a trailing slash. Resources can omit the trailing slash (such as a request with path parameters).

This is a [highly] [contentious] topic and that's just my opinion.

Internally, FastAPI registers both routes (with and without a trailing slash) and issues a 307 redirect to get to real endpoint.

[highly]: https://github.com/fastapi/fastapi/discussions/9328
[contentious]: https://github.com/fastapi/fastapi/pull/12145

app/v1/lifecycle/systems/endpoints.py
```diff
@@ -0,0 +1,42 @@
...
+):
+    systems = get_systems_data(major, minor)
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+def get_systems_data(major=None, minor=None):
+    data = OS_DATA_MOCKED
+
+    print(minor)
```
samdoran, line 35: This looks like a debugging statement that was left in.

hosekadam: Yeah, will remove it

hosekadam: Thanks for finding it :)

app/v1/lifecycle/systems/endpoints.py
```diff
@@ -0,0 +1,40 @@
...
+@v1_router.get("/systems")
+async def get_systems():
+    systems = get_systems_data()
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+@v1_router.get("/systems/{major}")
+async def get_systems_major(major: int = Path(..., description="Major version number")):
+    systems = get_systems_data(major)
```
samdoran, line 17: This is a call to a sync function from an async function. I need to look into this more, but I think this blocks the event loop.

I _think_ the correct way to call sync code from within async code is to use the default event loop with a Future object.

```suggestion
    loop = asyncio.get_running_loop()
    systems = await loop.run_in_executor(None, get_systems_data(major))
```

This isn't a problem currently because the blocking sync code is _fast_. But for a real IO or CPU bound operation, this could cause perceptible performance degradation.

app/v1/lifecycle/systems/endpoints.py
```diff
@@ -0,0 +1,40 @@
...
+
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+
+
+@v1_router.get("/systems/{major}/{minor}")
+async def get_systems_major_minor(
+    major: int = Path(..., description="Major version number"),
+    minor: int = Path(..., description="Minor version number"),
```
samdoran, line 25: A [better way to handle](https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/?h=annotated#better-with-annotated) this is to [use Annotated](https://fastapi.tiangolo.com/python-types/#type-hints-with-metadata-annotations).
```suggestion
    major: t.Annotated[int, Path(description="Major version number")],
    minor: t.Annotated[int, Path(description="Minor version number")],
```

It's also unnecessary to pass the ellipsis to `Path`.

PR #19, opened 2025-01-17
[RSPEED-422] Add app streams API

samdoran:
Add app streams lifecycle end point using fixture data.

The API uses path parameters for filtering and basic keyword filtering to find modules by name.

Example API queries with [httpie](https://httpie.io/docs/cli).
```
http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/
http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/ name==postgresql
http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/8
http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/9

http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/8/postgresql
http get localhost:8081/api/digital-roadmap/v1/lifecycle/app-streams/9/ruby
```

The shape of the returned data is up for discussion.

PR #20, opened 2025-01-21
Reorganize application code and tests

samdoran:
Simplify tests by using fixtures.
Reorganize modules and tests so the names line up and more clearly reflect their function.
Flatten modules so they have distinct names instead of the same names, such as `router.py` and `endpoints.py`.
Rename `app` to `roadmap` and move it under `src`.
Move to a source and tests organizational structure where the tests are outside of the source code and closely match the structure of the source code.

<details>
  <summary>Structure</summary>
  
```
src
└── roadmap
    ├── __init__.py
    ├── config.py
    ├── crud.py
    ├── data
    │   ├── __init__.py
    │   ├── modules.py
    │   ├── systems.py
    │   └── upcoming.py
    ├── database.py
    ├── main.py
    ├── models.py
    ├── services
    │   └── __init__.py
    └── v1
        ├── __init__.py
        ├── lifecycle
        │   ├── __init__.py
        │   ├── app_streams.py
        │   └── systems.py
        ├── release_notes.py
        └── upcoming.py
tests
├── __init__.py
├── conftest.py
├── test_main.py
└── v1
    ├── __init__.py
    ├── conftest.py
    ├── lifecycle
    │   ├── __init__.py
    │   ├── test_app_streams.py
    │   └── test_systems.py
    ├── test_release_notes.py
    └── test_upcoming.py
```

</details>


app/v1/lifecycle/systems/endpoints.py
hosekadam, line 1: Do we want to have the structure like `lifecycle/systems/endpoints.py` or `lifecycle/systems.py` as is introduced in #19 ? I think the PRs should be "compatible" and have the same structure - also in unit tests.

I this one (`lifecycle/systems/endpoints.py` or `lifecycle/app_streams/endpoints.py`) is better as we can have the code better organized if we have some specific functions/methods for each endpoint, and we don't want to have them in `endpoints.py`. Only disadvantage of this I see ther will be a lot of `endpoints.py` files

samdoran: I did a major reorganization of the repository. Let me know what you think.

hosekadam (dismissed):
Just one suggestion in `Makefile` and also I would remove the `/app` directory (as it's not used anymore). After doing so, I'm fine with changes and can be merged.

I agree that having a lot of files with the same name can be a bit confusing (reaction to my previous comment). I like how the repo looks now.

PR #28, opened 2025-01-24
Define response model for app streams

samdoran:
- Convert RHEL major version to a string
- Change format of returned dates
- Define app stream types
- Remove unnecessary default value from Field instances

### DIFF
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -1,12 +1,55 @@
-from pydantic import BaseModel, Field
+from datetime import date
+from enum import StrEnum
+
+from pydantic import BaseModel, Field, field_validator
+
+
+class AppStreamType(StrEnum):
+    module = "module"
+    package = "package"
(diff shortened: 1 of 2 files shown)


src/roadmap/models.py
```diff
@@ -1,12 +1,55 @@
...
+from datetime import date
+from enum import StrEnum
+
+from pydantic import BaseModel, Field, field_validator
+
+
+class AppStreamType(StrEnum):
+    module = "module"
+    package = "package"
+    scl = "scl"
```
hosekadam, line 10: I'm just curious: what is the `scl`?

hosekadam (approved):

src/roadmap/models.py
samdoran: That's the software collection.

samdoran:
Replaced by #47

PR #38, opened 2025-02-10
Reverse system order

hosekadam:
In the frontend is more useful to return the data in reverse order.

### DIFF
--- a/src/roadmap/v1/lifecycle/systems.py
+++ b/src/roadmap/v1/lifecycle/systems.py
@@ -1,3 +1,5 @@
+from operator import itemgetter
+
 from fastapi import APIRouter
 from fastapi import Path
 
@@ -14,14 +16,14 @@ router = APIRouter(
 async def get_systems():
     systems = get_systems_data()
 
-    return sorted(systems, key=lambda d: (d["major"], d["minor"]), reverse=True)
+    return sorted(systems, key=itemgetter("major", "minor"), reverse=True)
 
 
 @router.get("/{major}")
 async def get_systems_major(major: int = Path(..., description="Major version number")):
     systems = get_systems_data(major)
 
-    return sorted(systems, key=lambda d: (d["major"], d["minor"]), reverse=True)
+    return sorted(systems, key=itemgetter("major", "minor"), reverse=True)
 
 
 @router.get("/{major}/{minor}")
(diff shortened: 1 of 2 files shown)


samdoran:
This should be a parameter passed to the API so the caller can request the sort order. Up until now, we haven't implemented that but it's a very common thing and should exist in the API.

pr-watson:
> This should be a parameter passed to the API so the caller can request the sort order. Up until now, we haven't implemented that but it's a very common thing and should exist in the API.

For proper formatting for the lifecycle system data in the chart we're only going to need this to be in the reverse order so I don't think there would be much value in adding the parameter

src/roadmap/v1/lifecycle/systems.py
```diff
@@ -14,14 +14,14 @@
 async def get_systems():
     systems = get_systems_data()
 
-    return sorted(systems, key=lambda d: (d["major"], d["minor"]))
+    return sorted(systems, key=lambda d: (d["major"], d["minor"]), reverse=True)
```
samdoran, line 17: I generally try to avoid using `lamdba` in Python since they're pretty unreadable. I found a really nice way of doing this the other day using the [`operator` module](https://docs.python.org/3/library/operator.html#operator.attrgetter) and it does exactly what we need here.

> Return a callable object that fetches attr from its operand. If more than one attribute is requested, returns a tuple of attributes.

```suggestion
    from operator import attrgetter
    return sorted(systems, key=attrgetter("major", "minor"), reverse=True)
```

tests/v1/lifecycle/test_systems.py
```diff
@@ -51,4 +53,4 @@
 def test_system_specified(client, api_prefix, source_data, path, response, monkeypatch):
     monkeypatch.setattr(roadmap.v1.lifecycle.systems, "OS_DATA_MOCKED", source_data)
     data = client.get(f"{api_prefix}/lifecycle/systems{path}")
-    assert data.json() == response
+    assert data.json() == sorted(response, key=itemgetter("major", "minor"), reverse=True)
```
hosekadam, line 56: @samdoran Just thing to consider I found when I was changing the tests:

I was also about using the `sorted` for the response. But I was thinking that the code in the unit tests and in the code would be mostly simmilar and if there would be some mistake in the algorithm (the chance is low there, but we never know) we won't find it. So I rather changed the order of the expected order to have exactly what we want and don't touch the data in the test. I used for that some "reverse order of lines" tool, so it was pretty quick.

PR #39, opened 2025-02-10
[RSPEED-584] Return RHEL lifecycle data based on systems in inventory

samdoran:
[RSPEED-548](https://issues.redhat.com/browse/RSPEED-584)

Query the inventory API and return the count for each type of RHEL system as well as the appropriate lifecycle dates.

## Testing this branch ###

1. Get an access token: `export RH_TOKEN=$(scripts/get-redhat-access-token.py)`
1. Query the following API endpoints:

```
http localhost:8081/api/roadmap/v1/relevant/lifecycle/rhel/ Authorization:"Bearer $RH_TOKEN" 
http localhost:8081/api/roadmap/v1/relevant/lifecycle/rhel/9 Authorization:"Bearer $RH_TOKEN" 
http localhost:8081/api/roadmap/v1/relevant/lifecycle/rhel/9/1 Authorization:"Bearer $RH_TOKEN" 
```


scripts/scrub_results.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def main():
+    file = Path(sys.argv[1]).resolve()
+    data = json.loads(file.read_text())
+
+    new = {key: data[key] for key in set(data).difference(["results"])}
+    new["results"] = [{"system_profile": result["system_profile"]} for result in data["results"]]
+
+    file.with_suffix(".scrubbed.json").write_text(json.dumps(new, indent=2) + "\n")
```
beav, line 16: thank you for pretty printing this :)

src/roadmap/common.py
```diff
@@ -0,0 +1,93 @@
...
+
+# FIXME: This should be cached
+async def get_system_count_from_inventory(
+    headers: dict[str, str | None],
+    page: int = 1,
+    per_page: int = 100,
+    major: int | None = None,
+    minor: int | None = None,
+) -> dict[str, t.Any]:
+    if SETTINGS.dev:
```
beav, line 26: maybe break the contents of this if block into its own function for readability? It will make the "return" more obvious so folks know execution doesn't continue past the block. I had some trouble reading it

src/roadmap/v1/lifecycle/rhel.py
```diff
@@ -1,47 +1,185 @@
...
+        "User-Agent": user_agent,
+        "X-RH-Identity": x_rh_identity,
+    }
+    systems_response = await get_system_count_from_inventory(headers=headers, major=major, minor=minor)
+
+    system_counts = defaultdict(int)
+    for result in systems_response.get("results", []):
+        system_profile = result["system_profile"]
+        if not system_profile:
+            logger.info("Unable to get relevant systems due to missing system profile")
```
beav, line 100: if you have the system's ID here, print it so we are able to find the offending record

beav (approved):
one suggestion around logging but overall looks good for what it should do at this point.

pyproject.toml
```diff
@@ -37,6 +37,8 @@ addopts = [
     "--durations", "10",
     "--color", "yes",
     "--showlocals",
+    "--pdbcls", "IPython.terminal.debugger:TerminalPdb",
```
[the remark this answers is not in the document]
samdoran: I like having tab completion and `?` in the debugger.

src/roadmap/v1/lifecycle/rhel.py
samdoran: I put this in the log message for now. Ideally ID would be a separate field, but we'll have to do that when logging gets overhauled.

PR #43, opened 2025-02-19
Update upcoming and system mocked data

pr-watson:
This PR updates the systems data to have RHEL systems organized in descending order . The upcoming data has also been updated to have the proper naming for the `Addition` type and some updates to the first item in the array.


src/roadmap/data/upcoming.py
```diff
@@ -113,7 +113,7 @@
     },
     {
         "name": "MariaDB 11.0",
-        "type": "addition",
+        "type": "Addition",
```
samdoran, line 116: This seems like presentation that should be handled in the UI rather than data coming from the backend.

src/roadmap/data/systems.py
samdoran, line 1: Rather than changing the data, the code returning the response should be changed. That way it'll work consistently when we remove the mock data entirely.

pr-watson: Code returning the response has already been changed - https://github.com/RedHatInsights/digital-roadmap-backend/pull/38, this change just helps with the mocked data being used currently

src/roadmap/data/upcoming.py
pr-watson: I can change the frontend to also accept lowercase data but for consistency sake all the other `type` fields in the mocks are capitalized.

samdoran: Oh, I see what you mean. They should probably all be lowercase from the API and capitalized for consistency by the UI. The caller should assume it will need to formate the response data in some way. I don't think case is part of the schema.

samdoran (approved):

PR #44, opened 2025-02-19
Add tools for generating token and querying inventory

samdoran:
Update README and ignored files

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ API server providing access to Red Hat Enterprise Linux roadmap information.
 
 - Python 3.12 or later.
 - A container runtime such as `docker` or `podman`.
+- To use the `/relevant/` APIs, create an [offline token] in order to generate an access token.
 
 ### Prerequisites for `psycopg` ###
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -53,6 +54,32 @@ make start-db run
...
+
+```
+export RH_OFFLINE_TOKEN="[offline token]"
+export RH_TOKEN="$(./scripts/get-redhat-access-token.py)"
+```
+
+Use the access token in the request header. Here is an example using [httpie].
+
+```
+http localhost:8081/api/roadmap/v1/relevant/lifecycle/rhel/ \
```
hosekadam, line 80: This returns `404 Not found`, it's missing the commits from the branch with those changes. But I understand there is some "race" between merging those and relevant PR is still draft, so I don't see it as a problem now. I understand it as some preparation and when relevant get merged, it will work great.

(or we can use this for getting the token and try it in the relevant branch, I just think it's good to mention there :))

hosekadam (approved):
Approving, getting the token works great. I see this as a preparation for https://github.com/RedHatInsights/digital-roadmap-backend/pull/39 - now we are missing the `relevant` endpoint

README.md
samdoran: Correct. The changes in #39 will require using a token but nothing currently merged needs it.

PR #63, opened 2025-03-20
[RSPEED-828] Normalize app stream names

samdoran:
This feels like the wrong place to be doing this. It seems like we should be able to fetch the app stream names somehow instead of having to reconstruct them.

I think the challeng may be in needing to get the app stream name from the dnf module name.


beav:
@samdoran is it ok to close this, since the other PR is merged?

PR #69, opened 2025-03-25
[RSPEED-600] Calculate support status

hosekadam:
https://issues.redhat.com/browse/RSPEED-600

### DIFF
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -1,13 +1,13 @@
 import typing as t
 
-from datetime import date, timedelta
+from datetime import date
+from datetime import timedelta
 from enum import StrEnum
 
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 from pydantic import model_validator
-
 
 
 class Meta(BaseModel):
@@ -46,12 +46,11 @@ class System(BaseModel):
     minor: int | None = None
     release_date: date | t.Literal["Unknown"] | None
     retirement_date: date | t.Literal["Unknown"] | None
-    support_status: SupportStatus = SupportStatus.supported
+    support_status: SupportStatus = SupportStatus.unknown
     count: int = 0
     lifecycle_type: LifecycleType
 
-
-    @model_validator(mode='after')
+    @model_validator(mode="after")
     def update_support_status(self):
         """Validator for setting support status.
         Expected types/values of start_date and end_date are:
@@ -70,13 +69,13 @@ class System(BaseModel):
             if self.retirement_date < today:
                 self.support_status = SupportStatus.retired
                 return self
-            
+
             six_months_date = self.retirement_date - timedelta(days=180)
             if six_months_date <= today:
                 self.support_status = SupportStatus.six_months
             else:
                 self.support_status = SupportStatus.supported
-        
+
         else:
             self.support_status = SupportStatus.unknown
             return self
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -59,26 +59,9 @@ class System(BaseModel):
             - date(YYYY-MM-DD)
         """
         today = date.today()
-
-        if self.release_date != "Unknown" and self.release_date is not None:
-            if self.release_date > today:
-                self.support_status = SupportStatus.upcoming
-                return self
-
-        if self.retirement_date != "Unknown" and self.retirement_date is not None:
-            if self.retirement_date < today:
-                self.support_status = SupportStatus.retired
-                return self
-
-            six_months_date = self.retirement_date - timedelta(days=180)
-            if six_months_date <= today:
-                self.support_status = SupportStatus.six_months
-            else:
-                self.support_status = SupportStatus.supported
-
-        else:
-            self.support_status = SupportStatus.unknown
-            return self
+        self.support_status = _calculate_support_status(
+            start_date=self.release_date, end_date=self.retirement_date, current_date=today
+        )
 
         return self
 
@@ -107,3 +90,25 @@ class TaggedParagraph(BaseModel):
     title: str = Field(description="The paragraph title")
     text: str = Field(description="The paragraph text")
     tag: str = Field(description="The paragraph htmltag")
+
+
+def _calculate_support_status(
+    start_date: date | str | None, end_date: date | str | None, current_date: date
+) -> SupportStatus:
+    support_status = SupportStatus.unknown
+
+    if start_date not in (None, SupportStatus.unknown):
+        if start_date > current_date:
+            return SupportStatus.upcoming
+
+    if end_date not in (None, SupportStatus.unknown):
+        if end_date < current_date:
+            return SupportStatus.retired
+
+        six_months_date = end_date - timedelta(days=180)
+        if six_months_date <= current_date:
+            return SupportStatus.six_months
+
+        return SupportStatus.supported
+
+    return support_status
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -1,11 +1,13 @@
 import typing as t
 
-from datetime import date
+from datetime import date, timedelta
 from enum import StrEnum
 
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
+from pydantic import model_validator
+
 
 
 class Meta(BaseModel):
@@ -25,7 +27,8 @@ class SupportStatus(StrEnum):
     six_months = "Support ends within 6 months"
     retired = "Retired"
     not_installed = "Not installed"
-    upcoming = "Upcoming"
+    upcoming = "Upcoming release"
+    unknown = "Unknown"
 
 
 class HostCount(BaseModel):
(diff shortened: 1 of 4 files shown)


src/roadmap/models.py
```diff
@@ -48,6 +51,39 @@ class System(BaseModel):
...
+            if six_months_date <= today:
+                self.support_status = SupportStatus.six_months
+            else:
+                self.support_status = SupportStatus.supported
+        
+        else:
+            self.support_status = SupportStatus.unknown
+            return self
+
+        return self
```
beav, line 84: can this ever be hit? It looks like the "else" above would catch it

hosekadam: Yeah, that's right. Fixed, pushing also with the unit tests. Thanks!

samdoran:
You don't need to squash commits before merging. That's done when the branch is merged. It's actually nice to preserve the commit history in the PR branch.

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -153,3 +153,80 @@ def test_app_stream_package_single_digit():
...
+    mock_date = mocker.patch("roadmap.v1.lifecycle.app_streams.date")
+    mock_date.today.return_value = current_date
+    mock_date.side_effect = lambda *args, **kwargs: date(*args, **kwargs)
+
+    app_stream = AppStream(
+        name="pkg-name",
+        stream="1",
+        os_major=1,
+        os_minor=1,
+        os_lifecycle="mainline",
```
samdoran, line 224: This should be an enum not a string.
```suggestion
        os_lifecycle=LifecycleType.mainline,
```

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -153,3 +153,80 @@ def test_app_stream_package_single_digit():
...
+            SupportStatus.unknown,
+        ),
+    ),
+)
+def test_calculate_support_status_appstream(mocker, current_date, app_stream_start, app_stream_end, status):
+    # cannot mock the datetime.date.today directly as it's written in C
+    # https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking
+    mock_date = mocker.patch("roadmap.v1.lifecycle.app_streams.date")
+    mock_date.today.return_value = current_date
+    mock_date.side_effect = lambda *args, **kwargs: date(*args, **kwargs)
```
samdoran, line 217: A different way to do this is to wrap the specific value you want to mock. It avoids the use of `lambda` and is easier to read.
```suggestion
    mock_date = mocker.patch("roadmap.v1.lifecycle.app_streams.date", wraps=datetime.date)
    mock_date.today.return_value = current_date
```

tests/test_models.py
```diff
@@ -0,0 +1,80 @@
...
+    # https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking
+    mock_date = mocker.patch("roadmap.models.date")
+    mock_date.today.return_value = current_date
+    mock_date.side_effect = lambda *args, **kwargs: date(*args, **kwargs)
+
+    app_stream = System(
+        name="pkg-name",
+        major=9,
+        minor=6,
+        lifecycle_type="mainline",
```
samdoran, line 74:
```suggestion
        lifecycle_type=LifecycleType.mainline,
```

tests/test_models.py
```diff
@@ -0,0 +1,80 @@
...
+            SupportStatus.unknown,
+        ),
+    ),
+)
+def test_calculate_support_status_system(mocker, current_date, system_start, system_end, status):
+    # cannot mock the datetime.date.today directly as it's written in C
+    # https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking
+    mock_date = mocker.patch("roadmap.models.date")
+    mock_date.today.return_value = current_date
+    mock_date.side_effect = lambda *args, **kwargs: date(*args, **kwargs)
```
samdoran, line 68:
```suggestion
    mock_date = mocker.patch("roadmap.models.date", wraps=datetime.date)
    mock_date.today.return_value = current_date
```

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -153,3 +153,80 @@ def test_app_stream_package_single_digit():
...
+
+    app_stream = AppStream(
+        name="pkg-name",
+        stream="1",
+        os_major=1,
+        os_minor=1,
+        os_lifecycle="mainline",
+        count=4,
+        impl="package",
```
samdoran, line 226:
```suggestion
        impl=AppStreamImplementation.package,
```

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -139,6 +140,36 @@ def set_dates(self):  # noqa: C901
...
+        if self.end_date != "Unknown" and self.end_date is not None:
+            if self.end_date < today:
+                self.support_status = SupportStatus.retired
+                return self
+
+            six_months_date = self.end_date - timedelta(days=180)
+            if six_months_date <= today:
+                self.support_status = SupportStatus.six_months
+            else:
+                self.support_status = SupportStatus.supported
```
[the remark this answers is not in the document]
samdoran: You can see warnings in the test output.
```
tests/v1/lifecycle/test_app_streams.py::test_calculate_support_status_appstream[current_date5-None-app_stream_end5-Supported]
  /Users/sdoran/Developer/digital-roadmap-backend/tests/v1/lifecycle/test_app_streams.py:220: UserWarning: A custom validator is returning a value other than `self`.
  Returning anything other than `self` from a top level model validator isn't supported when validating via `__init__`.
  See the `model_validator` docs (https://docs.pydantic.dev/latest/concepts/validators/#model-validators) for more details.
    app_stream = AppStream(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

File: tests/v1/lifecycle/test_app_streams.py (hunk 3 above)
Comment (samdoran) [reply tests/v1/lifecycle/test_app_streams.py:217]: Another way to do this would be avoid mocking altogether and use dependency injection. That's a fancy way of saying add a `current_time` to the init and pass it in explicitly during testing. It does modify the code in order to test it, but that's not necessarily a bad thing.

```python
class AppStream(BaseModel):
    ...
    current_date: date = Field(default=date.today(), exclude=True)
```

Inside `update_support_status`, the `today` variable can be removed and the conditionals evaluate against the attribute:
```python
if self.start_date > self.current_date:
    ...
```

In the test
```python
def test_calculate_support_status_appstream(current_date, app_stream_start, app_stream_end, status):
    app_stream = AppStream(
        name="pkg-name",
        stream="1",
        os_major=1,
        os_minor=1,
        os_lifecycle=LifecycleType.mainline,
        count=4,
        impl=AppStreamImplementation.package,
        rolling=False,
        start_date=app_stream_start,
        end_date=app_stream_end,
        current_date=current_date,
    )

    assert app_stream.support_status == status
```

src/roadmap/models.py
```diff
@@ -43,10 +46,42 @@ class System(BaseModel):
...
+            if six_months_date <= today:
+                self.support_status = SupportStatus.six_months
+            else:
+                self.support_status = SupportStatus.supported
+
+        else:
+            self.support_status = SupportStatus.unknown
+            return self
+
+        return self
```
samdoran, line 83: Here's a simplified approach. 
```python
self.support_status = SupportStatus.unknown # This may be unnecessary but prevents return the passed in value.
if isinstance(self.start_date, date):
    if self.start_date > self.current_date:
        self.support_status = SupportStatus.upcoming
        return self

if isinstance(self.end_date, date):
    if self.end_date < self.current_date:
        self.support_status = SupportStatus.retired
        return self

    six_months_date = self.end_date - timedelta(days=180)
    if six_months_date <= self.current_date:
        self.support_status = SupportStatus.six_months
        return self

    self.support_status = SupportStatus.supported

return self
```

samdoran: If you wanted to pull this code out to use in both classes, you could turn it into a function and have this method call the function. It's not a big deal since it's only in two places, but that's an option.
```python
    @model_validator(mode="after")
    def update_support_status(self):
        return _validator_support_status(self)
```

src/roadmap/v1/lifecycle/app_streams.py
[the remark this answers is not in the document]
hosekadam: Okay, I'm adding it back thanks!

src/roadmap/models.py
hosekadam: It's needed: https://github.com/RedHatInsights/digital-roadmap-backend/pull/69#discussion_r2017541080 :D

(I removed it different file, in the models I forget to do it, but the code is mostly the same)

src/roadmap/models.py
hosekadam: That's exactly what I was thinking when I started applying the suggestions. "Why I haven't did this as a function and I need to change it twice?!"

I'm moving the status calculation into the `common.py`

Thanks for the review! :)

tests/v1/lifecycle/test_app_streams.py
hosekadam: didn't know about the wraps parameter, thanks

src/roadmap/models.py
samdoran: > So not in common.py there is an circular import...

We may need to rethink the organization a bit to avoid these issues. If we're getting enough utility functions, we could move those to a different module. Or avoid putting models in `common`.

> And instead of the isinstance() I'm using the not in (None, SupportStatus.unknown) expression. 

I switched to `isinstance` to make the type checker happy. I like the later expression better.

src/roadmap/models.py
```diff
@@ -72,3 +90,25 @@ class TaggedParagraph(BaseModel):
     title: str = Field(description="The paragraph title")
     text: str = Field(description="The paragraph text")
     tag: str = Field(description="The paragraph htmltag")
+
+
+def _calculate_support_status(
+    start_date: date | str | None, end_date: date | str | None, current_date: date
```
samdoran, line 96: This is a really nice refactor. It even makes the code easier to test because you can supply the date during testing (dependency injection).

tests/v1/lifecycle/test_app_streams.py
samdoran: Mocking is a dark art. There are an infinite number of things to learn and everyone will tell you you're doing it wrong.

samdoran (approved):

PR #79, opened 2025-03-31
add rbac url to env

beav:


beav:
@samdoran this says it passed CI and all checks have passed, but also says "Your project status has failed because the head coverage (97.52%) is below the target coverage (100.00%).". Is this ok to merge, or do I need to add tests?

PR #80, opened 2025-03-31
[RSPEED-914] Add job to sync inventory

samdoran:
Add a job than runs a script to sync inventory into our database.
Update the container build to create a new virtual environment to run the inventory sync script.
Freeze requirements for inventory sync script.

[RSPEED-914](https://issues.redhat.com/browse/RSPEED-914)


deploy/config.yml
```diff
@@ -137,3 +237,27 @@ parameters:
...
+    value: 250m
+
+  - name: CPU_LIMIT_INVENTORY_SYNC
+    value: 500m
+
+  - name: MEMORY_REQUEST_INVENTORY_SYNC
+    value: 256Mi
+
+  - name: MEMORY_LIMIT_INVENTORY_SYNC
+    value: 512Mi
```
beav, line 251: This seems too low but we can set it and see what happens. I would expect to need to bump it.

scripts/replication.py
```diff
@@ -0,0 +1,338 @@
+#!/usr/bin/env python
+# https://github.com/chambridge/replication_subscriber
```
beav, line 2: can you add the hash the file was pulled from? even if it's in a new line or something

Containerfile
```diff
@@ -50,6 +55,7 @@ RUN microdnf install -y --nodocs \
 RUN useradd --key HOME_MODE=0755 --system --create-home --home-dir /srv/roady roady
 
 COPY /src/roadmap/ /srv/roady/roadmap/
+COPY /scripts/replication.py /usr/local/bin/replication.py
```
beav, line 58: this is a clean way of doing things

deploy/config.yml
```diff
@@ -1,36 +1,39 @@
...
     kind: ClowdApp
     metadata:
-      name: "${APP_NAME}"
+      name: ${APP_NAME}
 
     spec:
       envName: ${ENV_NAME}
 
+      dependencies:
+        - rbac
```
beav, line 16: :+1:  thanks! i forgot about this

beav (dismissed):
I added some comments but none of them are blocking. lgtm

deploy/config.yml
samdoran: We can set a more sane default or override it later in the deployment config.

deploy/config.yml
```diff
@@ -1,36 +1,39 @@
...
+
       deployments:
-        - name: roadmap-api
+        - name: roadmap-api  # Would changing this to just API break 3scale?
```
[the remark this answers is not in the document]
samdoran: I think this is what's making the internal URL `roadmap-roadmap-api.foo`. I think it will be ok in the broader context but I don't know all the places it would show up. We also have a bunch of objects in OpenShift name `roadmap-roadmap-api` that look kinda goofy.

scripts/replication.py
samdoran: Oh, good idea.

PR #84, opened 2025-04-03
add psycopg2 dep

beav:


beav:
gonna self-ack this since its a dep update

PR #103, opened 2025-04-11
Stream results from the database

samdoran:
This should help improve responsiveness and performance.


src/roadmap/common.py
```diff
@@ -53,7 +53,8 @@ async def query_host_inventory(
                 if item.get("system_profile_facts", {}).get("operating_system", {}).get("minor") == minor
             ]
 
-        return response_data
+        for item in response_data:
+            yield item
```
beav, line 57: yay! this looks correct

beav (approved):

PR #105, opened 2025-04-14
[RSPEED-828] Add a display_name field to relevant app stream results

samdoran:
This is a field with normalized capitalization that handles many special cases and ensures a version number is present at the end.

Remove `stream` from the response schema since it is no longer available on the collated results.


beav (approved):

beav:
I'd like to see it with a couple of real-data cases I can think of, but this is good enough for sure to merge and see how it looks.

LGTM

samdoran:
We're using data from Stage for testing/development. We can add new data to see how it behaves pretty easily.

PR #107, opened 2025-04-15
[RSPED-1025] Use RBAC to check inventory access

samdoran:


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -214,8 +216,16 @@ async def get_relevant_app_streams(  # noqa: C901
     x_rh_identity: t.Annotated[str | None, Header(include_in_schema=False)] = None,
 ):
     org_id = decode_header(x_rh_identity)
-    # TODO: RBAC check. Is the requester allowed to access Inventory?
-    inventory_result = query_host_inventory(session=session, org_id=org_id)
+    rbac_response = await query_rbac(x_rh_identity)
```
beav, line 219: (just adding as reminder for us) this needs to be added to the rhel call too

samdoran: I already did that. :)

beav (approved):

PR #109, opened 2025-04-17
[RSPEED-1033] /upcoming-changes returns real data

Lifto:


src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+            detailFormat=0,
+            summary=upcoming.summary,
+            potentiallyAffectedSystems=5,
+            trainingTicket=upcoming.ticket,
+            dateAdded=date(2024, 10, 29),
+            lastModified=date(2024, 10, 29),
+        ),
+    )
+    for upcoming in upcomings
+]
```
samdoran, line 71: Building this as well as reading the file content should be encapsulated in a function. That will also allow [caching the result](https://fastapi.tiangolo.com/advanced/settings/?h=lru#read-settings-from-env).

```python
@cache
def get_upcoming_data():
    data = SETTINGS.upcoming_json_path.read_text()
    return [Upcoming.model_validate_json(item) for item in data]
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+
 
 @router.get("")
-async def get_upcoming():
-    # TODO: Replace fixture data with data from database
+async def get_upcoming() -> list[UpcomingResponse]:
```
samdoran, line 75: To align with other Red Hat APIs, the response model should include a `meta` field and data should be in `data` fields.

```python
class UpcomingResponse(BaseModel):
    meta: Meta
    data: list[Upcoming]

# In the route function
return {
    "meta": {"total": len(all_the_items), "count": len(items_on_this_page)},
    "data": get_upcoming_data()
}

File: src/roadmap/v1/upcoming.py (hunk 1 above)
Comment (samdoran) [reply src/roadmap/v1/upcoming.py:71]: If the JSON file matches the model, it can be passed directly to `model_validate()` or `model_validate_json()`, saving the work of translating the parameters. It might be worth reformatting the JSON to match the model.

File: tests/v1/test_upcoming.py [hunk 3]
```diff
@@ -1,3 +1,20 @@
...
+
+import roadmap.v1.upcoming
+
+
+def test_get_upcoming_changes(client, api_prefix, monkeypatch):
     response = client.get(f"{api_prefix}/upcoming-changes")
     assert response.status_code == 200
+    assert response.content.startswith(b'[{"name":"Node.js 22 included in RHEL 9 Application Streams","type":')
+
+    monkeypatch.setenv(
```
samdoran, line 13: This should be a separate test.

tests/v1/test_upcoming.py
```diff
@@ -1,3 +1,20 @@
...
+        "ROADMAP_UPCOMING_JSON_PATH",
+        str(Path(__file__).parent.parent.joinpath("fixtures").resolve().joinpath("upcoming.json")),
+    )
+    importlib.reload(roadmap.v1.upcoming)
```
samdoran, line 17: Restructuring the code so that the file is not read on import makes it easier to test and you won't have to reimport it.

samdoran (changes requested):

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+    summary: str
+    type: str
+    package: str
+    architecture: str
+    release: str
+    release_date: Date
+
+
+if not (filename := os.environ.get("ROADMAP_UPCOMING_JSON_PATH")):
+    filename = Path(__file__).parent.parent.joinpath("data").resolve().joinpath("upcoming.json")
```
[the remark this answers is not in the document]
Lifto: We have that here, yes?

Is it that we should enforce an absolute path if it is loaded from the environment? (googling implies we can check if it is, and we can make it absolute if need be.) Right now they could pass in an absolute or relative path in the env.

src/roadmap/v1/upcoming.py
Lifto: This would start with a change to data-pipeline-content, yes?

src/roadmap/v1/upcoming.py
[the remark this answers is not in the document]
Lifto: I'm thinking this is what we have, assuming the env var contains an absolute path.

Should we confirm it is an absolute path?

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+    title: str
+    summary: str
+    type: str
+    package: str
+    architecture: str
+    release: str
+    release_date: Date
+
+
+if not (filename := os.environ.get("ROADMAP_UPCOMING_JSON_PATH")):
```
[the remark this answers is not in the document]

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+    filename = Path(__file__).parent.parent.joinpath("data").resolve().joinpath("upcoming.json")
+
+with open(filename, "r") as file:
+    upcomings = TypeAdapter(list[Upcoming]).validate_json(file.read())
```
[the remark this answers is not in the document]

src/roadmap/v1/upcoming.py

tests/v1/test_upcoming.py
Lifto: fixed

tests/v1/test_upcoming.py
Lifto: fixed

src/roadmap/v1/upcoming.py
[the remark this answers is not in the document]
samdoran: The `.resolve()` method is what makes is absolute. I just meant you could omit that.

src/roadmap/v1/upcoming.py
samdoran: Potentially. We should make sure that the data here is inline with what's in `data-pipeline-content`. That's more authoritative and it's what we'll be dropping in during build.

src/roadmap/config.py
```diff
@@ -19,6 +22,7 @@ class Settings(BaseSettings):
     debug: bool = False
     dev: bool = False
     host_inventory_url: str = "https://console.redhat.com"
+    upcoming_json_path: FilePath = str(Path(__file__).parent.joinpath("data").resolve().joinpath("upcoming.json"))
```
samdoran, line 25: Casting this to a `str` removes all the nice functionality the `pathlib` provides.

Keep this path relative.
```suggestion
    upcoming_json_path: FilePath = Path(__file__).parent.joinpath("data").joinpath("upcoming.json")
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,81 @@
...
+
+
+class WrappedUpcomingResponse(BaseModel):
+    meta: Meta
+    data: list[UpcomingResponse]
+
+
+@cache
+def get_upcoming_data():
+    with open(Settings.create().upcoming_json_path, "r") as file:
```
samdoran, line 56: Use the already created Settings instance.
```suggestion
    with open(SETTINGS.upcoming_json_path, "r") as file:
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,81 @@
...
+        upcomings = TypeAdapter(list[Upcoming]).validate_json(file.read())
```
samdoran, line 57: What's the purpose of using a `TypeAdapter` here?

src/roadmap/v1/upcoming.py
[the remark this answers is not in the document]

src/roadmap/v1/upcoming.py
Lifto: Perhaps I should talk to the stakeholders, as I'm new to the party. 

In our thread I was told that the format we are returning from the API is the format they want, so I didn't consider changing it.

I take it `potentiallyAffectedSystems` is upcoming, and that we will cross-reference this with the list of hosts of the requesting user and tell them how many systems of theirs this applies to.

also data-pipeline-content is not yet, on develop anyway, providing added and last modified dates, so those will need to be real when they are available.

otherwise the only difference is the format. I would need to ask to find out who is concerned with the format data-pipeline-content is _providing_, maybe nobody will care if I change it. If so, I could change it to be the same as this one. (because I got the OK in a thread that API is making the frontend devs happy.)

I'm thinking changing this is beyond the scope of this update.

src/roadmap/config.py
Lifto: I did the `str` because pydantic was giving me a warning about an implicit conversion to string, but I can't reproduce the warning.

src/roadmap/v1/upcoming.py
Lifto: If I do this I won't be able to test it. The SETTINGS are made at import time, so the monkeypatch comes after. Suggestions?

I agree I don't like this usage.

src/roadmap/v1/upcoming.py
Lifto: The JSON is a list of objects, not an object. The TypeAdapter deserializes the list, and causes each of the elements of the list to be validated/deserialized via the Upcoming pydantic object.

tests/v1/test_upcoming.py
```diff
@@ -1,3 +1,24 @@
+from pathlib import Path
+
+import roadmap.v1.upcoming
+
+
 def test_get_upcoming_changes(client, api_prefix):
     response = client.get(f"{api_prefix}/upcoming-changes")
     assert response.status_code == 200
+    assert response.content.startswith(
+        b'{"meta":{"count":17,"total":17},"data":[{"name":"Node.js 22 included in RHEL 9 Application Streams","typ'
```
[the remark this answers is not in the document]
Lifto: fixed

Containerfile
```diff
@@ -57,6 +57,10 @@ RUN useradd --key HOME_MODE=0755 --system --create-home --home-dir /srv/roady ro
 COPY /src/roadmap/ /srv/roady/roadmap/
 COPY /scripts/replication.py /usr/local/bin/replication.py
 
+RUN curl -k \
+--output /srv/roady/roadmap/data/upcoming.json \
+"https://gitlab.cee.redhat.com/api/v4/projects/107966/repository/files/data%2F02_roadmap_jira.json/raw?ref=develop"
```
samdoran, line 62: nit: formatting and consistent use of long flag for clarity
```suggestion
RUN curl --insecure \
    --output /srv/roady/roadmap/data/upcoming.json \
    "https://gitlab.cee.redhat.com/api/v4/projects/107966/repository/files/data%2F02_roadmap_jira.json/raw?ref=develop"
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,81 @@
...
+    ]
+
 
 @router.get("")
-async def get_upcoming():
-    # TODO: Replace fixture data with data from database
-    return UPCOMING_DATA
+async def get_upcoming() -> WrappedUpcomingResponse:
+    data = get_upcoming_data()
+    return {"meta": {"total": len(data), "count": len(data)}, "data": data}
```
samdoran, line 81: nit: formatting for better legibility and easier to read diffs in the future. Trailing commas will cause `ruff` to format each item in multiple lines instead of folding.
```suggestion
    return {
        "meta": {
            "total": len(data),
            "count": len(data),
        },
        "data": data,
    }
```

src/roadmap/v1/upcoming.py
samdoran: Thanks for the explanation. I used this in another place but the docs didn't quite explain how it worked as well as you did.

src/roadmap/v1/upcoming.py
samdoran: But it sounds like we may hold off on changing the response format for now and maybe handle that in a follow up PR.

src/roadmap/v1/upcoming.py
samdoran: On solution would be to add an optional path parameter and use that if provided, using `SETTINGS.upcoming_json_path` by default. Call this function directly in testing instead of using a TestClient and provide a path to the settings file.

Probably the correct way to do this is to switch to using dependency injection for the settings so that it can be [overridden during testing](https://fastapi.tiangolo.com/advanced/testing-dependencies/?h=over#overriding-dependencies-during-testing).

src/roadmap/v1/upcoming.py
samdoran: > I take it potentiallyAffectedSystems is upcoming

Yes. We need to figure out how to get the list of system IDs and add it to the response. This will be a count of how many system IDs there are.

Looking at the current fixture data, which reflects thinking from very early on, and comparing it to `02_roadmap_jira.json`, I am wondering about the contents of the `details` field. I think we need a larger conversation around that because I'm not sure where most of that data is expected to come from.

>  data-pipeline-content is not yet, on develop anyway, providing added and last modified dates, so those will need to be real when they are available

Currently we're populating fields inside `details` with top level fields from the source data in `02_roadmap_jira.json`. That seems wrong. The data format should be more consistent end-to-end.

I think this is mainly an indicator that we should be trying to conform to the data format in `02_roadmap_jira.json` and not this original fixture data.

To your point, a broader discussion is needed with @pr-watson and @hosekadam. We can stick with the current solution for now.

samdoran:
This branch needs to be rebased in order to fix the merge conflict.

src/roadmap/v1/upcoming.py
samdoran: I did some more digging and experimentation. It's [not possible](https://github.com/fastapi/fastapi/discussions/10503#discussioncomment-7367434) to use FastAPIs dependency injection machinery outside of a route function. I figured that was the case.

We could switch injecting a settings instance and passing it around. That worked pretty well in my testing.

src/roadmap/v1/upcoming.py
Lifto: We got it updated, changes are in.

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,71 @@
...
+class Architecture(StrEnum):
+    arch64 = auto()
+    x86_64 = auto()
+    s390x = auto()
+    ppc64le = auto()
+    none = ""
+
+
+class UpcomingDetails(BaseModel):
+    architecture: Architecture | None
```
beav, line 40: the upcoming details will tailored off of packages first, not arch. However, better to have the un-tailored info returned and then tailor in a new PR.

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
+from datetime import date
+from enum import auto
+from enum import StrEnum
+from functools import lru_cache
+from pathlib import Path
+from typing import Annotated
+from typing import Any
```
samdoran, line 7: Just import `typing` once and then you can reference any type without having to do additional imports.
```suggestion
import typing as t
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+def get_upcoming_data(settings: Annotated[Settings, Depends(Settings.create)]) -> list[Upcoming]:
+    return read_upcoming_file(settings.upcoming_json_path)
+
 
 @router.get("")
-async def get_upcoming():
-    # TODO: Replace fixture data with data from database
-    return UPCOMING_DATA
+async def get_upcoming(data: Annotated[Any, Depends(get_upcoming_data)]) -> WrappedUpcoming:
+    return {"meta": {"total": len(data), "count": len(data)}, "data": data}
```
samdoran, line 76:
```suggestion
    return {
        "meta": {
            "total": len(data),
            "count": len(data),
        },
        "data": data,
    }
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+
+def get_upcoming_data(settings: Annotated[Settings, Depends(Settings.create)]) -> list[Upcoming]:
+    return read_upcoming_file(settings.upcoming_json_path)
+
 
 @router.get("")
-async def get_upcoming():
-    # TODO: Replace fixture data with data from database
-    return UPCOMING_DATA
+async def get_upcoming(data: Annotated[Any, Depends(get_upcoming_data)]) -> WrappedUpcoming:
```
samdoran, line 75:
```suggestion
async def get_upcoming(data: t.Annotated[list[Upcoming], Depends(get_upcoming_data)]) -> WrappedUpcoming:
```

src/roadmap/v1/upcoming.py
```diff
@@ -1,12 +1,76 @@
...
+    data: list[Upcoming]
+
+
+@lru_cache
+def read_upcoming_file(file: str | Path) -> list[Upcoming]:
+    with open(file, "r") as file:
+        return TypeAdapter(list[Upcoming]).validate_json(file.read())
+
+
+def get_upcoming_data(settings: Annotated[Settings, Depends(Settings.create)]) -> list[Upcoming]:
```
samdoran, line 70:
```suggestion
def get_upcoming_data(settings: t.Annotated[Settings, Depends(Settings.create)]) -> list[Upcoming]:
```

src/roadmap/v1/upcoming.py
Lifto: I'm sorry, I don't understand the request. Would you please provide more detail about the change you would like to see?

src/roadmap/v1/upcoming.py
Lifto: fixed

src/roadmap/v1/upcoming.py
Lifto: fixed

src/roadmap/v1/upcoming.py
Lifto: fixed

src/roadmap/v1/upcoming.py
Lifto: fixed

PR #118, opened 2025-04-22
Add tooling to load host inventory data for development and testing

samdoran:
This will create a new schema, `hbi`, and a new table `hosts` and load hosts from the fixture file. Then we are able to use that data is development and testing.

A new make target runs the loading program. It can be run many times and will remove all the existing host records before inserting again.

```
export ROADMAP_DEV=1
make start-db
make load-host-data
make run
```

The `/relevant/` APIs will query host inventory data.
```
http localhost:8000/api/roadmap/v1/relevant/lifecycle/app-streams
http localhost:8000/api/roadmap/v1/relevant/lifecycle/rhel
```

## Summary by Sourcery

Add tooling to load host inventory data for local development and testing environments

New Features:
- Create a new database schema and table to load host inventory data for local development
- Implement a script to populate the host inventory database with test data from fixtures

Enhancements:
- Remove hardcoded fixture data loading from query function
- Update README and Makefile to support loading host data during development

Build:
- Update development requirements files

Chores:
- Add a new Python script to manage host data loading
- Create a new database schema and host table for local testing

### DIFF
--- a/scripts/load_host_data.py
+++ b/scripts/load_host_data.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 
 import gzip
+import sys
 
 from datetime import datetime
 from datetime import timedelta
 from pathlib import Path
+from time import sleep
 from uuid import uuid4
 
 from app_common_python import json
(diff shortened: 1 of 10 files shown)


scripts/load_host_data.py
```diff
@@ -0,0 +1,120 @@
...
+                id=id,
+                display_name=f"{id.hex[:6]}.foo.redhat.com",
+                created_on=init_date,
+                modified_on=init_date,
+                system_profile_facts=host_data[n].get("system_profile_facts", {}),
+                ansible_host="ansible_host",
+                stale_timestamp=init_date + timedelta(30),
+                reporter="toast loader",
+                per_reporter_staleness={},
+                org_id="1234",
```
beav, line 105: can you change to 12345, so it matches the eph setup?

PR #120, opened 2025-04-22
Get database streaming working again

samdoran:
Go back to streaming the results from the database.

I needed to yield a single item from FastAPI's dependency injection machinery and do the `mappings()` call in the path function.

Depends on #118


beav (approved):

samdoran:
Tests are gonna fail until #118 is merged.

PR #122, opened 2025-04-23
Fix mocking org_id when ROADMAP_DEV is used

hosekadam:
## Summary by Sourcery

Bug Fixes:
- Ensure a consistent default org_id is set when in development mode, removing the conditional check

### DIFF
--- a/src/roadmap/common.py
+++ b/src/roadmap/common.py
@@ -112,8 +112,7 @@ async def query_host_inventory(
     minor: int | None = None,
 ):
     if settings.dev:
-        if not org_id:
-            org_id = "1234"
+        org_id = "1234"
 
     if groups:
         # TODO: Implement group filtering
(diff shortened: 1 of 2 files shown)


samdoran:
You'll need to update this test. https://github.com/RedHatInsights/digital-roadmap-backend/blob/1c0591ea2f0f1b80ef16341a1ad4d47a2b222b87/tests/test_common.py#L96

samdoran:
I was thinking it would be helpful to be able to override the `org_id` while in dev mode by manipulating the request header. But that's easy not to do when running the UI locally against the local API. Switching to always setting the `org_id` to match what's in the test data makes sense for that scenario.

If we ever need the ability to run in dev mode _and_ explicitly set the `org_id`, we can look at changing how the test data is loaded in the database.

samdoran (approved):

samdoran (approved):

PR #127, opened 2025-04-23
[RSPEED-775] Return system IDs for relevant systems

Lifto:
`relevant/lifecycle/app-streams` shows which systems are relevant to each app stream

### DIFF
--- a/src/roadmap/data/app_streams.py
+++ b/src/roadmap/data/app_streams.py
@@ -7766,9 +7766,7 @@ APP_STREAM_MODULES = [
     ),
 ]
 
-APP_STREAM_MODULES_BY_KEY = {
-    (asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES
-}
+APP_STREAM_MODULES_BY_KEY = {(asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES}
 
 OS_MAJORS_BY_APP_NAME = {}
 for asm in APP_STREAM_MODULES:
--- a/src/roadmap/data/app_streams.py
+++ b/src/roadmap/data/app_streams.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import typing as t
 
 from datetime import date
@@ -7768,13 +7769,9 @@ APP_STREAM_MODULES = [
 
 APP_STREAM_MODULES_BY_KEY = {(asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES}
 
-OS_MAJORS_BY_APP_NAME = {}
+OS_MAJORS_BY_APP_NAME = defaultdict(set)
 for asm in APP_STREAM_MODULES:
-    try:
-        major_versions = OS_MAJORS_BY_APP_NAME[asm.name]
-    except KeyError:
-        major_versions = set()
-        OS_MAJORS_BY_APP_NAME[asm.name] = major_versions
-    major_versions.add(asm.os_major)
+    OS_MAJORS_BY_APP_NAME[asm.name].add(asm.os_major)
+OS_MAJORS_BY_APP_NAME = dict(OS_MAJORS_BY_APP_NAME)
 
 APP_STREAM_MODULES_PACKAGES = [*APP_STREAM_MODULES, *APP_STREAM_PACKAGES.values()]
--- a/src/roadmap/data/app_streams.py
+++ b/src/roadmap/data/app_streams.py
@@ -7766,4 +7766,17 @@ APP_STREAM_MODULES = [
     ),
 ]
 
+APP_STREAM_MODULES_BY_KEY = {
+    (asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES
+}
+
+OS_MAJORS_BY_APP_NAME = {}
+for asm in APP_STREAM_MODULES:
+    try:
+        major_versions = OS_MAJORS_BY_APP_NAME[asm.name]
+    except KeyError:
+        major_versions = set()
+        OS_MAJORS_BY_APP_NAME[asm.name] = major_versions
+    major_versions.add(asm.os_major)
+
 APP_STREAM_MODULES_PACKAGES = [*APP_STREAM_MODULES, *APP_STREAM_PACKAGES.values()]
--- a/src/roadmap/data/app_streams.py
+++ b/src/roadmap/data/app_streams.py
@@ -1,6 +1,6 @@
-from collections import defaultdict
 import typing as t
 
+from collections import defaultdict
 from datetime import date
 from enum import StrEnum
 
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -40,11 +40,10 @@ RHELMajorVersion = t.Annotated[int, Path(description="Major RHEL version", ge=8,
 
 def get_rolling_value(name: str, stream: str, os_major: int) -> bool:
     try:
-        asm = APP_STREAM_MODULES_BY_KEY[(name, os_major, stream)]
+        return APP_STREAM_MODULES_BY_KEY[(name, os_major, stream)].rolling
     except KeyError:
         logger.debug(f"No match for rolling RHEL {os_major} {name} {stream}")
         return False
-    return asm.rolling
 
 
 def get_module_os_major_versions(name: str) -> set[int]:
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -7,7 +7,6 @@ from datetime import date
 from fastapi import APIRouter
 from fastapi import Header
 from fastapi import Path
-from fastapi.exceptions import HTTPException
 from fastapi.param_functions import Query
 from fastapi.params import Depends
 from pydantic import AfterValidator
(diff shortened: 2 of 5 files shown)


src/roadmap/data/app_streams.py
```diff
@@ -7766,4 +7766,15 @@ def set_os_version(self):
...
+APP_STREAM_MODULES_BY_KEY = {(asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES}
+
+OS_MAJORS_BY_APP_NAME = {}
+for asm in APP_STREAM_MODULES:
+    try:
+        major_versions = OS_MAJORS_BY_APP_NAME[asm.name]
+    except KeyError:
+        major_versions = set()
+        OS_MAJORS_BY_APP_NAME[asm.name] = major_versions
+    major_versions.add(asm.os_major)
```
samdoran, line 7778: Creating these different perspectives is a really good idea. It's also probably an indicator we need to rethink the data model, but that's for the future.

This can be simplified using a `defaultdict`
```suggestion
OS_MAJORS_BY_APP_NAME = defaultdict(set)
for asm in APP_STREAM_MODULES:
    OS_MAJORS_BY_APP_NAME[asm.name].add(asm.os_major)
```

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -37,22 +39,15 @@
...
 
 
 def get_module_os_major_versions(name: str) -> set[int]:
-    matches = set()
-    for item in APP_STREAM_MODULES:
-        if item.name == name:
-            matches.add(item.os_major)
-
-    return matches
+    return OS_MAJORS_BY_APP_NAME.get(name, set())
```
samdoran, line 50: This is so much better.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -207,23 +203,19 @@ async def get_app_stream_names(
 
 @relevant.get("", response_model=RelevantAppStreamsResponse)
 async def get_relevant_app_streams(  # noqa: C901
-    org_id: t.Annotated[str, Depends(decode_header)],
-    inventory_result: t.Annotated[t.Any, Depends(query_host_inventory)],
+    org_id: t.Annotated[str, Depends(decode_header)], systems: t.Annotated[t.Any, Depends(query_host_inventory)]
```
samdoran, line 206: Formatting nit:
```suggestion
    org_id: t.Annotated[str, Depends(decode_header)],
    systems: t.Annotated[t.Any, Depends(query_host_inventory)],
```

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -235,117 +227,128 @@ async def get_relevant_app_streams(  # noqa: C901
...
         missing_items = ", ".join(f"{key}: {value}" for key, value in missing.items())
         logger.info(f"Missing {missing_items} for org {org_id or 'UNKNOWN'}")
 
-    # Build response
     response = []
-    for count_key, count in module_count.items():
-        if count_key.rolling:
-            # Omit rolling app streams
+    for app_stream, systems in systems_by_stream.items():
+        if app_stream.rolling:
```
samdoran, line 250: Probably should leave the comment about it being purposefully omitted for now.

tests/test_config.py
```diff
@@ -46,7 +46,7 @@ def test_rbac_config_defaults(monkeypatch):
 
 def test_rbac_config_env(monkeypatch):
     monkeypatch.setenv("ROADMAP_RBAC_HOSTNAME", "example.com")
-    monkeypatch.setenv("ROADMAP_RBAC_PORT", 8080)
+    monkeypatch.setenv("ROADMAP_RBAC_PORT", "8080")
```
samdoran, line 49: Why does this need to be a string?

src/roadmap/data/app_streams.py
Lifto: I didn't make it a defaultdict because I thought it would make it prone to leaking if missing elements were requested.

Lifto: I would prefer to freeze it, but there don't know a lightweight way to do that.

src/roadmap/v1/lifecycle/app_streams.py
Lifto: lint won't allow it

src/roadmap/v1/lifecycle/app_streams.py
Lifto: fixed

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -235,117 +227,129 @@ async def get_relevant_app_streams(  # noqa: C901
...
+
+            app_key = AppStreamKey(
+                name=app_stream_package.application_stream_name,
+                application_stream_name=app_stream_package.application_stream_name,
+                stream=app_stream_package.stream,
+                start_date=app_stream_package.start_date,
+                end_date=app_stream_package.end_date,
+                os_major=os_major,
+                os_minor=os_minor if app_stream_package.rolling else None,
+                # TODO: Ask Brian if we want rolling releases to be displayed individually
```
beav, line 345: this TODO is complete if you want to remove it

Lifto: fixed

src/roadmap/data/app_streams.py
Lifto: @r0x0d I'm curious how it would help here. I read up on dataclasses just now, it seems it can freeze the assignment of a field, but I don't see how it can freeze adding to a dictionary or set.

samdoran: > I didn't make it a defaultdict because I thought it would make it prone to leaking if missing elements were requested.

Ah, I see what you mean. You could change it to a `dict`. Doing all this at the module global level feels kinda weird, but I guess it works.

```suggestion
OS_MAJORS_BY_APP_NAME = defaultdict(set)
for asm in APP_STREAM_MODULES:
    OS_MAJORS_BY_APP_NAME[asm.name].add(asm.os_major)
OS_MAJORS_BY_APP_NAME = dict(OS_MAJORS_BY_APP_NAME)
```

samdoran: A dataclass (or any class) could work by raising an `AttributeError` instead setting a default value when accessing an attribute that doesn't exist.

We could dynamically create a frozen Pydantic model to house the data, but we would have to use `getattr` for items that have a `-` in them.

```python
OSMajorAppName = create_model("OSMajorAppName", **{asm.name: (set[int], ...) for asm in APP_STREAM_MODULES})```

File: src/roadmap/v1/lifecycle/app_streams.py (hunk 3 above)
Comment (samdoran) [reply src/roadmap/v1/lifecycle/app_streams.py:206]: A [comma after the last argument](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma) will do the trick.

File: src/roadmap/data/app_streams.py (hunk 1 above)
Comment (samdoran) [reply src/roadmap/data/app_streams.py:7778]: If we wanted to keep the global namespace a bit cleaner, we could make function and call that.

```python
def get_os_majors_by_app_name(values: list[AppStreamEntity]) -> dict[str, set[int]]:
    result = defaultdict(set)
    for asm in values:
        result[asm.name].add(asm.os_major)

    return dict(result)

OS_MAJORS_BY_APP_NAME = get_os_majors_by_app_name(APP_STREAM_MODULES)
```

This isn't a blocker, just musing over some different ways to do things.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -235,117 +227,125 @@ async def get_relevant_app_streams(  # noqa: C901
...
         "meta": {
-            "count": len(module_count),
+            "count": len(response),
             "total": sum(item.count for item in response),
         },
         "data": sorted(response, key=sort_attrs("name", "os_major", "os_minor", "os_lifecycle")),
     }
+
+
+def app_streams_from_modules(dnf_modules, os_major, os_minor, os_lifecycle):
```
samdoran, line 283: Add type hints.

samdoran:
The only other thing to address is test coverage is a bit low and the lint ignores can be removed. I need to enable the rule to check for unnecessary `noqa` statements.

src/roadmap/v1/lifecycle/app_streams.py
Lifto: fixed

src/roadmap/data/app_streams.py
Lifto: I made a change here. I hear you on the choices available.

Lifto:
> The only other thing to address is test coverage is a bit low and the lint ignores can be removed. I need to enable the rule to check for unnecessary `noqa` statements.

I made a Jira ticket for testing. I didn't want to hold up this improvement on the tests, which would be a small project itself.

I tried taking out the `noqa` statements but they still cause C901s.  I'd be happy to refactor that code, but I think we gotta get this one out.

src/roadmap/v1/lifecycle/app_streams.py
Lifto: fixed

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -235,117 +227,125 @@ async def get_relevant_app_streams(  # noqa: C901
...
+            os_lifecycle=os_lifecycle if rolling else None,
+            rolling=rolling,
+            impl=AppStreamImplementation.module,
+        )
+        app_streams.add(app_key)
+
+    return app_streams
+
+
+def app_streams_from_packages(package_names_string, os_major, os_minor, os_lifecycle):
```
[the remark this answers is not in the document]
Lifto: fixed

samdoran:
I'll take care of the linter settings in #129.

We're under pressure to deliver, so I'm ok merging this with less than ideal test coverage. Once we're past this storm, we should require test coverage before merging changes.

samdoran (approved):

PR #132, opened 2025-04-24
[RSPEED-723] Return related application stream info

Lifto:


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -263,30 +246,23 @@ async def get_relevant_app_streams(
...
-                if app.end_date is None or app.end_date > date.today():
-                    relateds[(app.display_name, app.os_major)] = app
-    for app_stream in app_streams:
-        try:
-            del relateds[(app_stream.name, app_stream.os_major, app_stream.display_name)]
-        except KeyError:
-            pass
-    return relateds.values()
+            if app.name == app_stream.name:
+                if streams_lt(app_stream.stream, app.stream):
```
samdoran, line 259: You could define `__lt__` methods on the classes in order to do the comparison here.

```python
    def __lt__(self, comparison: str):
        mine = [int(part) for part in self.stream.split('.')]
        return mine < [int(part) for part in comparison.split('.')]
```

It could have better error handling to make it more robust, but that's the gist.

src/roadmap/common.py
```diff
@@ -186,3 +186,18 @@ def ensure_date(value: str | date):
...
+
+
+def _split_stream(stream: str):
+    """Returns a tuple of major, minor and micro for given stream."""
+    versions = stream.split('.')
+    versions.reverse()
+    major = versions.pop()
+    minor = versions.pop() if versions else ""
+    micro = versions.pop() if versions else ""
+    return (major, minor, micro)
```
samdoran, line 198: These probably need to be `int` in order to compare properly. Otherwise it's doing a lexical comparison which may not always be correct.

Lifto: good catch, thank you

src/roadmap/v1/lifecycle/app_streams.py
Lifto: I thought about this, but it seemed too loose to define __lt__ on AppStreamEntity for just this (though maybe I should reconsider), and too heavy to make a Stream class and give it an __lt__ method. I do grok the __lt__ method, but maybe I'm not seeing where you'd put it.

src/roadmap/common.py
Lifto: fixed. Not, a few items have a stream of "rhel8"

beav (approved):

src/roadmap/v1/lifecycle/app_streams.py
samdoran: It would go on the `AppStreamEntity` class. But maybe it's too clever to compare an object to a string.
```python
if app_stream < app.stream:
...
```

src/roadmap/common.py
```diff
@@ -186,3 +186,21 @@ def ensure_date(value: str | date):
         return date.fromisoformat(value)
     except (ValueError, TypeError):
         raise ValueError("Date must be in ISO 8601 format")
+
+
+def _split_stream(stream: str):
```
samdoran, line 191:
```suggestion
def _split_stream(stream: str) -> tupe[int, int, int]:
```

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -357,54 +325,21 @@ def app_streams_from_modules(
...
     return app_streams
 
 
 def app_streams_from_packages(
     package_names_string: str,
     os_major: str,
-    os_minor: str,
-    os_lifecycle: str,
-) -> set[AppStreamKey]:
+) -> set[AppStreamEntity]:
```
samdoran, line 336: It looks like defining `__hash__` and only putting in the keys we want to collect on allowed us to get rid of AppStreamKey. That's pretty sweet.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -250,10 +178,88 @@ async def get_app_stream_names(
 @relevant.get("", response_model=RelevantAppStreamsResponse)
 async def get_relevant_app_streams(
     org_id: t.Annotated[str, Depends(decode_header)],
-    systems: t.Annotated[t.Any, Depends(query_host_inventory)],
+    systems: t.Annotated[AsyncResult, Depends(query_host_inventory)],
+    related: bool = False,
 ):
     logger.info(f"Getting relevant app streams for {org_id or 'UNKNOWN'}")
 
+    systems_by_stream = await systems_by_app_stream(systems, org_id)
```
samdoran, line 186: This should be an dependency since `systems` isn't needed directly.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -250,10 +178,88 @@ async def get_app_stream_names(
 @relevant.get("", response_model=RelevantAppStreamsResponse)
 async def get_relevant_app_streams(
     org_id: t.Annotated[str, Depends(decode_header)],
-    systems: t.Annotated[t.Any, Depends(query_host_inventory)],
+    systems: t.Annotated[AsyncResult, Depends(query_host_inventory)],
```
samdoran, line 181:
```suggestion
    systems_by_stream: t.Annotated[dict[AppStreamEntity, list[UUID]], Depends(systems_by_app_stream)],
```

src/roadmap/common.py
Lifto: fixed

samdoran:
It looks like there are now more entries for the same display_name than before. This will most likely look incorrect in the UI.

<details>
  <summary>Apache HTTPD 2.4 entries in the response</summary>
  
```
> http \
    --pretty format \
    localhost:8000/api/roadmap/v1/relevant/lifecycle/app-streams | \
    jello '[(item.name, item.display_name, item.os_major, item.start_date) for item in _.data if "httpd" in item.name.casefold()]' | bat -l yaml
STDIN
  [
    [
      "httpd",
      "Apache HTTPD 2.4",
      8,
      "2019-05-07"
    ],
    [
      "httpd",
      "Apache HTTPD 2.4",
      9,
      "2022-05-18"
    ],
    [
      "httpd-core",
      "Apache HTTPD 2.4",
      9,
      "2022-11-15"
    ],
    [
      "httpd-filesystem",
      "Apache HTTPD 2.4",
      9,
      "2022-05-18"
    ],
    [
      "httpd-tools",
      "Apache HTTPD 2.4",
      9,
      "2022-05-18"
    ]
  ]
```
</details>

<details>
  <summary>Current output from <code>main</code></summary>
  
```
> http \
    --pretty format \
    localhost:8000/api/roadmap/v1/relevant/lifecycle/app-streams | \
    jello '[(item.name, item.display_name, item.os_major, item.start_date) for item in _.data if "httpd" in item.name.casefold()]' | bat -l yaml

  [
    [
      "Apache httpd 2.4",
      "Apache HTTPD 2.4",
      9,
      "2022-05-18"
    ],
    [
      "Apache httpd 2.4",
      "Apache HTTPD 2.4",
      9,
      "2022-11-15"
    ],
    [
      "httpd",
      "Apache HTTPD 2.4",
      8,
      "2019-05-07"
    ]
  ]
```
</details>

src/roadmap/v1/lifecycle/app_streams.py
Lifto: I changed the arguments to `systems_by_app_stream` to be dependencies, is that what you meant?

```
async def systems_by_app_stream(systems: t.Annotated[AsyncResult, Depends(query_host_inventory)],
) -> dict[AppStreamEntity, list[UUID]]:
```

Lifto: oh, now I'm looking at the next comment, you meant something else.

Lifto:
About the additional streams we see in the new output, I'm not sure which is better, the old or the new.

Looking into it, what is happening is that on main, we are creating an AppStreamKey (which, btw, bc it is a pydantic object, gets a `__hash__`` for free based on all of its pydantic fields) and we are assigning `name` to be `application_stream_name`, and `name` does not appear in the AppStreamEntity object. However on this branch, we are using `name` (as well as `display_name` and `application_stream_name`) in `__hash__`.

looking at the data, It seems these `name`s all map to the same `application_stream_name`, Apache HTTPD 2.4

httpd-core
httpd-filesystem
httpd-tools
httpd

So, on the branch, you see all four of those, but on main those four have collapsed into one.

I'm not sure which is the better response, main or branch.

one strike against main is that it has two identically named packages + os versions, with different dates. One is httpd, and one is https-core.

samdoran:
I think the results here are more correct but we should double check how it's handled by the UI. If the UI consolidates items further by `display_name` and sums the counts, then it may be ok. Otherwise I think we may have to look at counting based on `display_name` and omitting the other name fields. That may have some odd ramifications to consider.

samdoran:
The UI doesn't do any special handling or consolidation of the results — they are displayed as returned. That means the burden is on the API to consolidate the records for presentation. That separation of concerns is debatable, but for now we probably need to group based on `display_name`.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -240,18 +168,47 @@ async def get_app_stream_names(
...
+def related_app_streams(app_streams: list[AppStreamKey]) -> list[AppStreamKey]:
+    """Return unique list of related apps that do not appear in app_streams."""
+    relateds = set()
+    for app_stream_key in app_streams:
+        for app in APP_STREAM_MODULES_PACKAGES:
+            if app.match_name == app_stream_key.app_stream_entity.match_name:
+                if streams_lt(app_stream_key.app_stream_entity.stream, app.stream):
+                    if app.end_date is None or app.end_date > date.today():
+                        relateds.add(AppStreamKey(app, app_stream_key.name))
+    return relateds - set(app_streams)
```
abadger, line 204:
```suggestion
    return relateds.difference(app_streams)
```
Saves on converting the `app_streams` list into a set. (In general, I like to avoid using operator syntax for sets too because they aren't self-documenting.)

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -240,18 +168,47 @@ async def get_app_stream_names(
...
+        return isinstance(other, AppStreamKey) and self.__hash__() == other.__hash__()
+
+
+def related_app_streams(app_streams: list[AppStreamKey]) -> list[AppStreamKey]:
+    """Return unique list of related apps that do not appear in app_streams."""
+    relateds = set()
+    for app_stream_key in app_streams:
+        for app in APP_STREAM_MODULES_PACKAGES:
+            if app.match_name == app_stream_key.app_stream_entity.match_name:
+                if streams_lt(app_stream_key.app_stream_entity.stream, app.stream):
```
abadger, line 201: Would we want to handle the equality case as well?  I'm not sure, for instance, if having the same version but a different `end_date` might occur and be important information?

abadger (commented):
Took a look from a superficial perspective.  A few cosmetic-y things and two things related to `streams_lt()` that might be cornercases to address.

src/roadmap/common.py
```diff
@@ -186,3 +186,21 @@ def ensure_date(value: str | date):
...
+    micro = int(versions.pop()) if versions else 0
+    return (major, minor, micro)
+
+
+def streams_lt(a: str, b: str):
+    """Return True if stream a is less than stream b."""
+    try:
+        return _split_stream(a) < _split_stream(b)
+    except ValueError:
+        return a < b
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/data/app_streams.py
```diff
@@ -77,6 +80,72 @@ def set_os_version(self):
...
+            "mysql": "MySQL",
+            "nginx": "NGINX",
+            "node.js": "Node.js",
+            "nodejs": "Node.js",
+            "openjdk": "OpenJDK",
+            "osinfo-db": "OSInfo DB",
+            "php": "PHP",
+            "postgresql": "PostgreSQL",
+            "rhn-tools": "RHN Tools",
+        }
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/data/app_streams.py
```diff
@@ -77,6 +80,72 @@ def set_os_version(self):
...
+            self.name,
+            self.display_name,
+            self.application_stream_name,
+            self.os_major,
+            self.os_minor,
+            self.start_date,
+            self.end_date,
+            self.impl,
+            self.rolling,
+        ).__hash__()
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -240,18 +168,47 @@ async def get_app_stream_names(
     }
 
 
-## Relevant ##
-relevant = APIRouter(
-    prefix="/relevant/lifecycle/app-streams",
-    tags=["Relevant", "App Streams"],
-)
+class AppStreamKey:
+    """Wraps AppStreamEntity objects in"""
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/v1/lifecycle/app_streams.py
Lifto: fixed. (thanks, didn't know this one!)

src/roadmap/v1/lifecycle/app_streams.py
Lifto: I can't agree or disagree here. We've been having discussions and it's a fine concoction we seek. I'm going to leave this unchanged and resolve this, but I am about to go review a recent discussion with Brian Smith and John Carr where we get into some of these details. It seems easy to provide too much information, and such changes would cause us to return more results. They may be technically correct, but cluttering of the UI. I'm deferring to more product and design situated people for what we want here. However, you are correct there could be such matches of stream with differing dates, and there probably is a better choice to make here.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -357,54 +278,118 @@ def app_streams_from_modules(
...
+                    related=False,
+                )
             )
-            app_streams.add(app_key)
+        except Exception as exc:
+            raise HTTPException(detail=str(exc), status_code=400)
 
-    return app_streams
+    if related:
+        for app_stream in related_app_streams(systems_by_stream.keys()):
```
samdoran, line 340: This is passing in a dict_keys object when the function is expecting a list. Maybe the type hint just needs updating to `t.KeysView[AppStreamKey]`.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -357,54 +278,118 @@ def app_streams_from_modules(
...
+                        count=0,
+                        rolling=app_stream.app_stream_entity.rolling,
+                        systems=[],
+                        related=True,
+                    )
+                )
+            except Exception as exc:
+                raise HTTPException(detail=str(exc), status_code=400)
+
+    if len(relevant_app_streams) == 0:
```
samdoran: There may be a reason for the explicit equality check. If not, it's more idiomatic to use a falsy check.
```suggestion
    if not relevant_app_streams:
```

samdoran: Is this the section that is adding some results if no related items were found? Probably worth putting a comment to explain that.

This should probably be handled in a separate PR since the behavior could be seen as surprising. It's worthy of a design discussion.

src/roadmap/v1/lifecycle/rhel.py
```diff
@@ -134,9 +138,51 @@ async def get_relevant_systems(
...
+                    minor=os.minor,
+                    lifecycle_type=LifecycleType.mainline,
+                    release_date=os.start,
+                    retirement_date=os.end,
+                    count=0,
+                    related=True,
+                )
+            )
+
+    if len(results) == 0:
```
samdoran: Same comments as above.
```suggestion
    if not results:
```

samdoran (changes requested):

src/roadmap/data/app_streams.py
```diff
@@ -28,6 +45,8 @@ class AppStreamEntity(BaseModel):
     """An application stream module or package."""
 
     name: str = Field(min_length=1)
+    display_name: str | None = None
+    match_name: str | None = None
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -240,18 +168,49 @@ async def get_app_stream_names(
     }
 
 
-## Relevant ##
-relevant = APIRouter(
-    prefix="/relevant/lifecycle/app-streams",
-    tags=["Relevant", "App Streams"],
-)
+class AppStreamKey:
```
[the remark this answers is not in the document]
Lifto: I didn't see why it would help to make it a pydantic model when I needed to override the hash function, though I could have looked to see if it would do a nested hash well. (notice it holds a reference to an AppStreamEntity, and constructs its hash using the contents of same) 

It does seem in this code base we opt for using pydantic models over plain classes. I can change it and see how it works.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -240,18 +168,49 @@ async def get_app_stream_names(
...
-@relevant.get("", response_model=RelevantAppStreamsResponse)
-async def get_relevant_app_streams(
+
+def related_app_streams(app_streams: list[AppStreamKey]) -> set[AppStreamKey]:
+    """Return unique list of related apps that do not appear in app_streams."""
+    relateds = set()
+    for app_stream_key in app_streams:
+        for app in APP_STREAM_MODULES_PACKAGES:
+            if app.match_name == app_stream_key.app_stream_entity.match_name:
+                if streams_lt(app_stream_key.app_stream_entity.stream, app.stream):
```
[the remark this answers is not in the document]
Lifto: I'll give it a try and look at some results.

src/roadmap/v1/lifecycle/app_streams.py
Lifto: ah, I didn't actually know the type of dict.keys()

src/roadmap/v1/lifecycle/app_streams.py
Lifto: this is on another branch now.

src/roadmap/v1/lifecycle/app_streams.py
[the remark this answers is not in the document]
Lifto: made the change, works

src/roadmap/v1/lifecycle/app_streams.py
Lifto: how about `Iterable`, as that's the only real concern for the input. (DictKeys is overconstrainted)

Lifto: updated

src/roadmap/v1/lifecycle/app_streams.py
Lifto: fixed on other branch

src/roadmap/v1/lifecycle/rhel.py
Lifto: fixed on other branch

samdoran (approved):

PR #138, opened 2025-04-29
Reference workspaaces instead of groups in error message

samdoran:
The user facing term is now “workspaces” instead of “groups”.


tests/test_common.py
```diff
@@ -173,19 +187,20 @@ async def test_query_rbac_no_url():
 
 
 async def test_check_inventory_access():
-    result = await check_inventory_access([{"resourceDefinitions": [], "permission": "inventory:*:*"}])
+    perms = [{"resourceDefinitions": [], "permission": "inventory:*:*"}]
+    result = await check_inventory_access(perms)
 
-    assert result == []
+    assert result == perms
```
samdoran, line 193: I'm not sure this test is providing much utility now that it just returns the passed in value so long as the permissions match.

beav (approved):

PR #142, opened 2025-04-30
[RSPEED-752] Return core related info when no systems present

Lifto:
note: branched off of https://github.com/RedHatInsights/digital-roadmap-backend/pull/132, merge that first.

### DIFF
--- a/src/roadmap/common.py
+++ b/src/roadmap/common.py
@@ -203,4 +203,6 @@ def streams_lt(a: str, b: str):
     try:
         return _split_stream(a) < _split_stream(b)
     except ValueError:
-        return a < b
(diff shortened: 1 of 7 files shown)


Lifto:
There is more than one way to approach this. I opted to automatically return some pre-determined apps and oses when a query would produce no results. Another option would be to add a flag, perhaps core=true, that would mean "include core relateds", but I would think this would make the UI trickier, as the UI would have to load, see there were no (or too few) results, and then re-load, this time with the core=true flag. That or always have core=true, which defeats the purpose of having a flag. So I opted to have the API automatically return these pre-determined relateds when it was going to otherwise return nothing.

PR #147, opened 2025-05-07
add streaming=parallel to subscription

beav:


samdoran:
We are going to move to using the read-only replica since we were unable to get streaming replication to work in production.

PR #159, opened 2025-05-13
packages must match stream major to be in results

Lifto:


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -289,20 +289,30 @@ def app_streams_from_modules(
...
+
+
 def app_streams_from_packages(
-    package_names_string: str,
+    package_names_string: list[str],
     os_major: str,
 ) -> set[AppStreamKey]:
-    package_names = {pkg.split(":")[0].rsplit("-", 1)[0] for pkg in package_names_string}
-
+    packages = [StringPackage.from_string(s) for s in package_names_string]
```
samdoran, line 308: Should this be a set to avoid duplicates?

Lifto: fixed

PR #162, opened 2025-05-14
[RSPEED-1400] Potentially affected systems

Lifto:


src/roadmap/v1/upcoming.py
```diff
@@ -38,37 +42,88 @@ class Architecture(StrEnum):
...
 
 class WrappedUpcoming(BaseModel):
     meta: Meta
-    data: list[Upcoming]
+    data: list[UpcomingOutput]
 
 
 @lru_cache
-def read_upcoming_file(file: str | Path) -> list[Upcoming]:
+def read_upcoming_file(file: str | Path) -> list[UpcomingInput]:
```
samdoran, line 88: This should only accept a `Path` object since that's what is being provided by settings.

src/roadmap/v1/upcoming.py
```diff
@@ -38,37 +42,88 @@ class Architecture(StrEnum):
...
     with open(file, "r") as file:
-        return TypeAdapter(list[Upcoming]).validate_json(file.read())
-
-
-def get_upcoming_data(settings: t.Annotated[Settings, Depends(Settings.create)]) -> list[Upcoming]:
-    return read_upcoming_file(settings.upcoming_json_path)
+        return TypeAdapter(list[UpcomingInput]).validate_json(file.read())
```
samdoran, line 90: If this is a `Path` object, this can be simplified a little.
```python
return TypeAdapter(list[UpcomingInput]).validate_json(file.read_text())
```

src/roadmap/v1/upcoming.py

src/roadmap/v1/upcoming.py
```diff
@@ -38,37 +42,87 @@ class Architecture(StrEnum):
...
+        for key in keys_by_name[upcoming.package]:
+            systems.update(systems_by_stream[key])
+
+        details = UpcomingOutputDetails(
+            architecture=upcoming.details.architecture,
+            detailFormat=upcoming.details.detailFormat,
+            summary=upcoming.details.summary,
+            trainingTicket=upcoming.details.trainingTicket,
+            dateAdded=upcoming.details.dateAdded,
+            lastModified=upcoming.details.lastModified,
```
samdoran, line 111: This looks like `upcoming.details` could just be passed and `UpcomingOutput` should do something with it , or just reuse the existing values instead of typing them all out.
```suggestion
            **upcoming.details.model_dump(),
```

Lifto: line 112 and 113 add more data. I didn't know a fancier pattern for doing what I was trying to do here.

src/roadmap/v1/upcoming.py
```diff
@@ -38,37 +42,87 @@ class Architecture(StrEnum):
...
+            potentiallyAffectedSystemsCount=len(systems),
+            potentiallyAffectedSystems=systems,
+        )
+
+        result.append(
+            UpcomingOutput(
+                name=upcoming.name,
+                type=upcoming.type,
+                release=upcoming.release,
+                date=upcoming.date,
```
samdoran, line 121: Probably can do something similar here as well but you have conflicting `details` attributes.

samdoran (approved):
Let's merge this because deadlines are upon us. When we have some breathing room, we should scrutinize the data structure because it feel like we're doing gymnastics now instead of the data just flowing smoothly.

samdoran (approved):

PR #166, opened 2025-05-15
Add /relevant/upcoming-changes API

Lifto:
* new endpoint: /relevant/upcoming-changes
Preston requested a version of the /upcoming-changes endpoint that only returned changes that pertained to hosts of the user. /relevant/upcoming-changes does that. 

* package (name) now included in results of all upcoming-changes endpoints, per Preston's request:
  > We do check inventory against that package field so it would make sense to have it in the modal description


samdoran:
Can we get a description of the goal of this change?

Lifto:
> Can we get a description of the goal of this change?

description added (see top comment)

pr-watson (approved):
LGTM

samdoran (commented):
Do we want to change `/api/roadmap/v1/upcoming-changes` to return data _without_ having the query inventory now that we're adding this API? If someone is only interested in getting all the upcoming changes and has no hosts in inventory, we should be able to provide just the data.

src/roadmap/v1/upcoming.py
```diff
@@ -137,3 +139,21 @@ async def get_upcoming(data: t.Annotated[t.Any, Depends(get_upcoming_data)]) ->
...
     }
+
+
+relevant = APIRouter(
+    prefix="/relevant/upcoming-changes",
+    tags=["Relevant", "Upcoming Changes"],
+)
+
+
+@relevant.get("")
```
samdoran, line 150: This API needs a summary and maybe a description to explain what's different and how it works. Summary is a parameter like the API above. Description can either be a parameter or the doctoring on the function.

Lifto:
> Do we want to change `/api/roadmap/v1/upcoming-changes` to return data _without_ having to query inventory now that we're adding this API? If someone is only interested in getting all the upcoming changes and has no hosts in inventory, we should be able to provide just the data.

Considering that the only consumer of this API is Roadmap UI, and Roadmap UI has not requested that feature, my opinion is that we should not add it. I'm for adding it if they (or any new consumer) requests it. It is a sensible feature.

Lifto:
> The API is not only meant to be consumed by the UI. There are customers that want to get roadmap data that do not have hosts in inventory (same for RHEL and app stream lifecycle data).

in this case we should have such an endpoint.

I'm asking Preston if I can just take hosts out of `/upcoming-changes` or if we need a new plan

Lifto:
Proposing:
/upcoming-changes  --  no host data
/relevant/upcoming-changes   --  only changes for your hosts, w host data
/relevent/upcoming-changes?all=true   --  all changes, w host data

src/roadmap/v1/upcoming.py

pr-watson (approved):

tests/v1/test_upcoming.py
```diff
@@ -6,6 +6,28 @@
 
 
 def test_get_upcoming_changes(client, api_prefix):
+    response = client.get(f"{api_prefix}/upcoming-changes")
+    assert response.status_code == 200
+    data = response.json()["data"]
+    assert data[0]["name"] == "Add Node.js to RHEL9 AppStream THIS IS TEST DATA"
+    for d in data:
+        assert "potentiallyAffectedSystems" not in d["details"]
+        assert "potentiallyAffectedSystemsCount" not in d["details"]
```
samdoran, line 15: Could of nits:

Group setup and assertions instead of intermingling them to make tests easier to read.

Instead of `for` loops, try to break up assertions into statements with explanations. It will make figuring out the purpose of the test and why it failed easier in the future.
```suggestion
    response = client.get(f"{api_prefix}/upcoming-changes")
    data = response.json()["data"]
    
    assert response.status_code == 200
    assert data[0]["name"] == "Add Node.js to RHEL9 AppStream THIS IS TEST DATA"
    assert not any("potentiallyAffectedSystems" in d["details"] for d in data), "Why did this fail?"
    assert not any("potentiallyAffectedSystemsCount" in d["details"] for d in data), "Why did this fail?"
```

src/roadmap/v1/upcoming.py
```diff
@@ -74,22 +74,46 @@ class UpcomingOutputDetails(BaseModel):
...
+    return {
+        "meta": {
+            "total": len(data),
+            "count": len(data),
+        },
+        "data": data,
+    }
+
+
+def get_upcoming_data_with_hosts(
```
samdoran, line 116: nit: Put the routes at the bottom of the file and supporting classes/functions at the top of the file.

samdoran (approved):

tests/v1/test_upcoming.py
Lifto: fixed

PR #172, opened 2025-05-15
[RSPEED-1454] Improve database engine creation and make pool size configurable

samdoran:
The database _engine_ should be created only once and reused for the life of the application. The _session_ should be opened and closed for each API request.

Add two new settings for controlling the connection pool size, `ROADMAP_DB_POOL_SIZE`, and the maximum number of connections that can be created, `ROADMAP_DB_MAX_OVERFLOW`. The total number of connections is `pool_size` + `max_overflow`. After that, SQLAlchemy will raise errors.

https://fastapi.tiangolo.com/tutorial/sql-databases/
https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r

### DIFF
--- a/src/roadmap/database.py
+++ b/src/roadmap/database.py
@@ -1,14 +1,22 @@
-import typing as t
-
-from fastapi import Depends
 from sqlalchemy.ext.asyncio import async_sessionmaker
 from sqlalchemy.ext.asyncio import create_async_engine
 
 from roadmap.config import Settings
 
 
-async def get_db(settings: t.Annotated[Settings, Depends(Settings.create)]):
-    engine = create_async_engine(str(settings.database_url), echo=True)
-    async_session = async_sessionmaker(engine, expire_on_commit=False)
+SETTINGS = Settings.create()
+ENGINE = create_async_engine(
+    str(SETTINGS.database_url),
+    echo=True,
+    pool_size=SETTINGS.db_pool_size,
+    max_overflow=SETTINGS.db_max_overflow,
+)
+
+
+async def get_db():
+    # Docs for connections and transactions that explain pool_size and max_overflow
+    #   https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r
+
+    async_session = async_sessionmaker(ENGINE, expire_on_commit=False)
     async with async_session() as session:
         yield session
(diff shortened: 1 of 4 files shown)


src/roadmap/database.py
```diff
@@ -1,14 +1,22 @@
...
+    pool_size=SETTINGS.db_pool_size,
+    max_overflow=SETTINGS.db_max_overflow,
+)
+
+
+async def get_db():
+    # Docs for connections and transactions that explain pool_size and max_overflow
+    #   https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r
+
+    async_session = async_sessionmaker(ENGINE, expire_on_commit=False)
```
Lifto, line 20: If we want the preserve the dependency injection pattern for Settings

```suggestion

ENGINE = None

def get_engine(settings: t.Annotated[Settings, Depends(Settings.create)]):
    # do we have to say "global ENGINE" here maybe?
    if ENGINE is None:
        ENGINE = create_async_engine(
        str(settings.database_url),
        echo=True,
        pool_size=settings.db_pool_size,
        max_overflow=settings.db_max_overflow,
    )
    return ENGINE
    
async def get_db(engine: Depends(get_engine)):
     # Docs for connections and transactions that explain pool_size and max_overflow
     #   https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r

     async_session = async_sessionmaker(engine, expire_on_commit=False)
```

Lifto: maybe sticking with dependency injection preserves (perhaps among other things) our ability to change these settings in testing, though I don't think we change these settings in testing..

Lifto (commented):
left a comment. I don't think it's necessary to make the change, but thought I'd leave a comment.

src/roadmap/database.py
samdoran: I tried that and the problem is it has the same problem of creating one engine per API request. It needs to be created on import so that it's done once for the lifecycle of the application.

This might make it harder to test in the future, but we use the real database connection during testing and don't currently have a need to override the settings here.

I looked into using [lifespan events](https://fastapi.tiangolo.com/advanced/events/) for this but this was the example from the docs and seems to be the most straightforward.

Side note, you would have to use `global ENGINE` in order for that work as intended, but using globals is generally a bad idea.

samdoran: It's possible we could cache the result or create a class to manage the database connection that caches the engine.

PR #174, opened 2025-05-16
Ignore non-host RBAC permissions

Lifto:
We were rejecting requests when RBAC was returning permissions like this:

```[{'permission': 'inventory:hosts:write', 'resourceDefinitions': []},
 {'permission': 'inventory:hosts:read', 'resourceDefinitions': []},
 {'permission': 'inventory:groups:write', 'resourceDefinitions': []},
 {'permission': 'inventory:groups:read', 'resourceDefinitions': []},
 {'permission': 'inventory:groups:read',
  'resourceDefinitions': [{'attributeFilter': {'key': 'group.id',
                                               'operation': 'in',
                                               'value': ['c22abc43-62f9-4a03-94e0-2a49d0e3c3d8']}}]}]
```

the populated resourceDefinition was in inventory:groups, when our query only pertains to inventory:hosts.

this PR has our code filter out permissions that do not pertain to hosts before analyzing them. This prevents an RBAC result like the one above from causing a 501.

### DIFF
--- a/src/roadmap/common.py
+++ b/src/roadmap/common.py
@@ -96,16 +96,13 @@ async def check_inventory_access(
 
     Return list of resource definitions and permission.
     """
-    inventory_access_perms = {"inventory:*:*", "inventory:hosts:read"}
+    inventory_access_perms = {"inventory:*:*", "inventory:*:read", "inventory:hosts:read"}
+    host_permissions = [p for p in permissions if p.get("permission") in inventory_access_perms]
 
-    has_access = False
-    if any(permission.get("permission") in inventory_access_perms for permission in permissions):
-        has_access = True
-
-    if not has_access:
+    if not host_permissions:
         raise HTTPException(status_code=403, detail="Not authorized to access host inventory")
 
-    return permissions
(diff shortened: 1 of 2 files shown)


beav:
we had a big convo about this on slack. looks good to me.

beav (dismissed):

samdoran:
Can we add tests for this?

samdoran:
Can some details of that conversation be shared here for context? Otherwise this is a change to RBAC handling with little to no information.

Lifto:
I'll whip up a test

Lifto:
Test added. Confirmed this test fails on branch main

samdoran:
That makes sense. The original intent and current behavior is to return a 501 for _any_ resource definitions. But it seems like what we should do is return a 501 only for resource definitions related to inventory access since. Resource definitions unrelated to inventory access are effectively ignored

We need to process inventory-related resource definitions and adjust the query based accordingly, but that will need to be done in a follow up PR.

samdoran (approved):

PR #175, opened 2025-05-19
Properly parse package names that are missing the expected delimiter

samdoran:
Some packages may not have a `:`. Use the partition method to avoid a ValueError and partition on `-` instead.

### DIFF
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -311,7 +311,12 @@ class StringPackage(BaseModel, frozen=True):
 
     @classmethod
     def from_string(cls, s):
-        name, major = s.split(":")
+        name, _, major = s.partition(":")
+        if not major:
+            # Missing ':' in the expected package name. Partition on '-' instead.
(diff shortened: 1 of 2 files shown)


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -311,7 +311,12 @@ class StringPackage(BaseModel, frozen=True):
 
     @classmethod
     def from_string(cls, s):
-        name, major = s.split(":")
+        name, _, major = s.partition(":")
+        if not major:
```
Lifto, line 315: Would it not be more in the spirit of `partition` to check `_` for `""` instead of checking `major` ?

Lifto (commented):
Looks good to me. My comment is just a nit.

src/roadmap/v1/lifecycle/app_streams.py
samdoran: That might read a little better. It's equivalent since both the last two items will both be empty strings if the delimiter isn't found.

samdoran: I don't really like using `_`.

PR #176, opened 2025-05-20
[RSPEED-1107] RBAC groups

Lifto:
Honor RBAC permissions that limit host access to certain groups.

in order to test run `make load-host-data`, as the test data has been updated.


src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
...
+        elif resource_definition["attributeFilter"].get("operation") != "in":
+            raise HTTPException(501, detail="Invalid value for attributeFilter.operation in RBAC response.")
+        elif not isinstance(resource_definition["attributeFilter"]["value"], list):
+            raise HTTPException(501, detail="Did not receive a list for attributeFilter.value in RBAC response.")
+        else:
+            # Validate that all values in the filter are UUIDs.
+            group_list = resource_definition["attributeFilter"]["value"]
+            try:
+                for gid in group_list:
+                    if gid is not None:
```
beav, line 104: when would gid be none?

src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
...
 
     if not host_permissions:
         raise HTTPException(status_code=403, detail="Not authorized to access host inventory")
 
-    return host_permissions
+    for perm in host_permissions:
+        # Get the list of allowed Group IDs from the attribute filter.
+        for resourceDefinition in perm["resourceDefinitions"]:
+            if len(resourceDefinition) == 0:
+                # Any record with an empty resourceDefinition means
```
beav, line 141: Can you add a reference in the comment to a doc or code snippet w/ this rule?

src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
...
+    host_groups: t.Annotated[list[dict], Depends(get_allowed_host_groups)],
     major: MajorVersion = None,
     minor: MinorVersion = None,
 ):
     if settings.dev:
         org_id = "1234"
 
-    if any(perm.get("resourceDefinitions") for perm in permissions):
-        logger.info(f"RBAC {permissions}")
-        # TODO: Implement workspace filtering
```
beav, line 122: from TO-DO to TO-DONE :)

src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
     return data.get("data", [{}])
 
 
-async def check_inventory_access(
+def _get_group_list_from_resource_definition(resource_definition: dict) -> list[str]:
+    if "attributeFilter" in resource_definition:
+        if resource_definition["attributeFilter"].get("key") != "group.id":
+            raise HTTPException(501, detail="Invalid value for attributeFilter.key in RBAC response.")
```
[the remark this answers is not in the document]
Lifto: not that I know of. I'm cribbing this code from https://github.com/RedHatInsights/insights-host-inventory/blob/9736d60a16fbc4af219aef844cc2485bd12af408/lib/middleware.py#L161

They seemed to think an "in" was the only condition worth checking for

src/roadmap/common.py
Lifto: I don't know. This code is very defensive so as to not imply unrestricted access. The check for None is copied from line 149 in this: https://github.com/RedHatInsights/insights-host-inventory/blob/9736d60a16fbc4af219aef844cc2485bd12af408/lib/middleware.py#L161

src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
...
     if major is not None:
         query = f"{query} AND system_profile_facts #>> '{{operating_system,major}}' = :major"
 
     if minor is not None:
         query = f"{query} AND system_profile_facts #>> '{{operating_system,minor}}' = :minor"
 
+    if host_groups:
+        string_ids = [f"'{u}'" for u in host_groups]
+        id_string = ",".join(string_ids)
+        query = f"{query} AND EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE group_obj->>'id' IN ({id_string}));"
```
beav, line 172: This query took me a few min to figure out. Can you add a comment?

Also I think the semicolon at the end is not needed

src/roadmap/common.py
```diff
@@ -87,48 +88,89 @@ async def query_rbac(
...
+            raise HTTPException(501, detail="Did not receive a list for attributeFilter.value in RBAC response.")
+        else:
+            # Validate that all values in the filter are UUIDs.
+            group_list = resource_definition["attributeFilter"]["value"]
+            try:
+                for gid in group_list:
+                    if gid is not None:
+                        UUID(gid)
+            except (ValueError, TypeError):
+                raise HTTPException(501, detail="Received invalid UUIDs for attributeFilter.value in RBAC response.")
```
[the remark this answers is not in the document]
Lifto: piping the org_id through the code is going to be ugly.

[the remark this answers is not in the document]
Lifto: logging added for the bad UUID (not for org_id)

src/roadmap/common.py
Lifto: I am having a hard time finding any such reference, I will keep digging.

src/roadmap/common.py
Lifto: 🎉

src/roadmap/common.py
Lifto: Sam found it:
https://insights-rbac.readthedocs.io/en/latest/management.html#resource-definitions

"In general, ALL roles should be created with a resourceDefinitions stanza of []. This is taken to mean “no additional filtering” and will generally result in the expected behavior.

In specific cases where the application logic has been written to handle them, however, resource access can be limited by specifing an attribute filter in the resourceDefinitions stanza as below."

src/roadmap/common.py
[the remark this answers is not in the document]
samdoran: Looks like these are the allowed values. https://github.com/RedHatInsights/insights-rbac/blob/e3ab728c89de1ec9cc75fb51060b352b9bd357a6/rbac/management/role/serializer.py#L37

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
     return data.get("data", [{}])
 
 
-async def check_inventory_access(
+def _get_group_list_from_resource_definition(resource_definition: dict) -> list[str]:
+    if "attributeFilter" in resource_definition:
+        if resource_definition["attributeFilter"].get("key") != "group.id":
```
samdoran, line 93: Do the lookup once and store the result. Doing the lookup multiple times is somewhat expensive in Python.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
+            raise HTTPException(501, detail="Invalid value for attributeFilter.key in RBAC response.")
+        elif resource_definition["attributeFilter"].get("operation") != "in":
+            raise HTTPException(501, detail="Invalid value for attributeFilter.operation in RBAC response.")
+        elif not isinstance(resource_definition["attributeFilter"]["value"], list):
```
samdoran, line 97: Is `value` guaranteed to exist or should this use the `get()` method like the other values?

samdoran: Why is this an instance check? This almost seems like we're validating the response, not evaluating the value. We should be able to trust the response and only be concerned about the data.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
+            raise HTTPException(501, detail="Did not receive a list for attributeFilter.value in RBAC response.")
+        else:
```
samdoran, line 99: This `else` isn't necessary.

src/roadmap/common.py
samdoran: That explains the `[None]` I'm seeing currently. I thought that looked funny.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
+        #     "ungrouped": false,
+        #     "updated": "2025-01-07T12:59:52.471612+00:00"
+        #    }
+        # ]
+        # This addition to the query will query into that field's list of group
+        # info. It searches for any record with an id that matches any of our
+        # eligible host group ids.
+        string_ids = [f"'{u}'" for u in host_groups]
+        id_string = ",".join(string_ids)
+        query = f"{query} AND EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE group_obj->>'id' IN ({id_string}))"
```
samdoran, line 191: Using f-strings to build a SQL query is asking for a SQL injection attack. The list of IDs needs to be a token and passed in as a param.

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -241,6 +239,38 @@ async def decode_header_override():
...
+    async def query_rbac_override():
+        return [
+            {
+                "permission": "inventory:hosts:read",
+                "resourceDefinitions": [
+                    {
+                        "attributeFilter": {
+                            "key": "group.id",
+                            "operation": "in",
+                            "value": ["aec18a86-3593-11f0-8426-5e43c8b8aa2f", "397e1696-34f2-11f0-a718-5e43c8b8aa2f"],
```
samdoran, line 252: It would be good add a test case for `[None]` as well since that's what we're see most often right now.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
+    # an attributeFilter. Ensure downstream usage of this result is not
+    # interpreted as unrestricted access!
+    return []
+
+
+async def get_allowed_host_groups(
     permissions: t.Annotated[list[dict[t.Any, t.Any]], Depends(query_rbac)],
-) -> list[dict]:
+) -> t.Iterable[UUID]:
     """Check the given permissions for inventory access.
```
samdoran, line 122: This doctoring needs updating now that the behavior has changed.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
 
     Raise HTTPException if no permissions allow access.
 
-    Return list of resource definitions and permission.
+    Return list of groups hosts may belong to, if restricted, otherwise returns
+    an empty list, meaning unrestricted access to the org's hosts.
```
samdoran, line 127: This will return a list or set. The doc or the function should be updated.

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
+            if len(resourceDefinition) == 0:
+                # Any record with an empty resourceDefinition means
+                # unrestricted access.
+                # https://insights-rbac.readthedocs.io/en/latest/management.html#resource-definitions
+                return []
+
+            group_list = _get_group_list_from_resource_definition(resourceDefinition)
+            allowed_group_ids.update(group_list)
+
+    return allowed_group_ids
```
samdoran, line 150:
```suggestion
    return list(allowed_group_ids)
```

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
 
 
-# FIXME: This should be cached
 async def query_host_inventory(
     org_id: t.Annotated[str, Depends(decode_header)],
     session: t.Annotated[AsyncSession, Depends(get_db)],
     settings: t.Annotated[Settings, Depends(Settings.create)],
-    permissions: t.Annotated[list[dict], Depends(check_inventory_access)],
+    host_groups: t.Annotated[list[dict], Depends(get_allowed_host_groups)],
```
samdoran, line 157:
```suggestion
    host_groups: t.Annotated[list[UUID], Depends(get_allowed_host_groups)],
```

src/roadmap/common.py
Lifto: comment added

src/roadmap/common.py
samdoran: I think SQLalchemy may do a lot of work for you.
```suggestion
        
        query = f"{query} AND EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE group_obj->>'id' IN :host_groups)"
    result = await session.stream(
         text(query),
         params={
            "org_id": org_id,
            "major": str(major),
            "minor": str(minor),
            "host_groups: host_groups,
        },
    )
```

src/roadmap/common.py
```diff
@@ -87,48 +88,108 @@ async def query_rbac(
...
     host_permissions = [p for p in permissions if p.get("permission") in inventory_access_perms]
 
     if not host_permissions:
         raise HTTPException(status_code=403, detail="Not authorized to access host inventory")
 
-    return host_permissions
+    for perm in host_permissions:
+        # Get the list of allowed Group IDs from the attribute filter.
+        for resourceDefinition in perm["resourceDefinitions"]:
+            if len(resourceDefinition) == 0:
```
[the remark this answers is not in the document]
Lifto: fixed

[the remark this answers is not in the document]
Lifto: I took it back. While I'm not going to check the type, a value of "False" would be wrong here. It is supposed to be a `[]` for us to interpret it as "unrestricted access"

src/roadmap/common.py
samdoran: I think this query might need some work. Rather than taking the first item, we should aggregate all the `id` values from within the list of dicts in `groups`. That's a massive pain with SQL but it is possible.

I don't think grabbing one item from `groups` is correct.

src/roadmap/common.py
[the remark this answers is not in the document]
samdoran: `if not []` evaluates to `True`.

```
In [65]: not []
Out[65]: True

In [66]: not [1]
Out[66]: False

In [68]: not [] == (len([]) == 0)
Out[68]: True
```

src/roadmap/common.py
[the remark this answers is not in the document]
Lifto: added in code for equal, needs tests still

src/roadmap/common.py
Lifto: I put that code in because in the case that they do send us malformed data, we'll get a clearer picture of what is wrong. Sometimes you get weird data! (like the [None] issue https://issues.redhat.com/browse/RHINENG-16842. It could even be spec to send us unexpected data. By validating our own expectations we prevent granting unrestricted access accidentally. Sometimes services give altered or invalid responses, and some defensive programming on our part prevents propagating or escalating those invalid responses.

src/roadmap/common.py
Lifto: fixed

src/roadmap/common.py
samdoran: I reworked this query to use a common table expression (CTE) that creates a list of all the group IDs then only returns records where any of those values are in the list of `host_groups`.
```python
async def query_host_inventory(
    session: t.Annotated[AsyncSession, Depends(get_db)],
    settings: t.Annotated[Settings, Depends(Settings.create)],
    host_groups: t.Annotated[list[UUID], Depends(get_allowed_host_groups)],
    major: MajorVersion = None,
    minor: MinorVersion = None,
):
    if settings.dev:
        org_id = "1234"

    query_str = "SELECT * FROM hbi.hosts WHERE org_id = :org_id"
    if host_groups:
        query_str = """
        WITH hosts_with_group_ids AS (
            SELECT *,
                (
                    SELECT ARRAY_AGG(element->>'id')
                    FROM json_array_elements(groups) AS element
                    WHERE json_array_length(groups) > 0
                ) as group_ids
            FROM hbi.hosts
            WHERE json_array_length(groups) > 0
        )
        SELECT * from hosts_with_group_ids
        WHERE group_ids && :host_groups
        """
    if major is not None:
        query_str = f"{query_str} AND system_profile_facts #>> '{{operating_system,major}}' = :major"

    if minor is not None:
        query_str = f"{query_str} AND system_profile_facts #>> '{{operating_system,minor}}' = :minor"

    query = text(textwrap.dedent(query_str))
    result = await session.stream(
        query,
        params={
            "org_id": org_id,
            "major": str(major),
            "minor": str(minor),
            "host_groups": host_groups,
        },
    )
    yield result
```

samdoran: It would probably be _much_ better to move the filters for major and minor versions inside the CTE.

src/roadmap/common.py
```diff
@@ -87,55 +89,131 @@ async def query_rbac(
...
+                # TODO: Check if this None check can be removed once the Kessel Phase 0 migration is complete.
+                # https://issues.redhat.com/browse/RHINENG-16842
+                # Notes update:
+                # I have a conversation from #team-insights-inventory that explains
+                # that we should treat a None as if it were the ID of a group called
+                # "ungrouped", and it is identified by the field 'ungrouped: true'
+                # in the group data.
+                if gid is None:
+                    raise NotImplementedError
+                if gid is not None:
```
beav, line 123: should this be an `else`? I'm assuming either lines 121 or 123 have to be true

src/roadmap/common.py
beav: it is grabbing one item? I read it as "if the system is in any group, return 1 from the clause", which the 'exists' then checks

src/roadmap/common.py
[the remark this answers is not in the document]
Lifto: test is in

src/roadmap/common.py
Lifto: fixed

src/roadmap/common.py
Lifto: I took it out. I hope RBAC is reliable.

src/roadmap/common.py
Lifto: I didn't see this, but I made a change and some tests. Please have a second look and tell me if we should change this. 

As to the "select 1", it's like Beav is saying, it's the same as an "any" in Python.

tests/v1/lifecycle/test_app_streams.py
Lifto: test for this and related cases are in

src/roadmap/common.py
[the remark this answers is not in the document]
Lifto: I put it back to just an truthy check with 'if'. I figure if RBAC sends us garbage, and all the cows get out, they can be responsible.

src/roadmap/common.py
Lifto: fixed

src/roadmap/common.py
Lifto: I return a set because the downstream users of this will be checking for membership.

It gets converted to a list before it goes into the SQL query.

src/roadmap/common.py
Lifto: I changed this to str. (str | None, actually) because we don't hold it as a type UUID, we just use UUID to confirm it the string is a legal UUID

src/roadmap/common.py
Lifto: fixed

Lifto:
I should have marked this draft until now, pardon.

src/roadmap/common.py
```diff
@@ -87,55 +89,133 @@ async def query_rbac(
...
+        #     "name": "GroupTwo",
+        #     "org_id": "1234",
+        #     "ungrouped": false,
+        #     "updated": "2025-01-07T12:59:52.471612+00:00"
+        #    }
+        # ]
+
+        # There is a special case. If None is in host_groups, we must query
+        # for a group that has ungrouped == True.
+        ungrouped_query = "EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE (group_obj->>'ungrouped')::boolean = true)"
```
beav, line 199: just confirming, 'ungrouped' is there today in the database? or does it have to wait for HBI to push something

src/roadmap/common.py
```diff
@@ -87,55 +89,133 @@ async def query_rbac(
...
+
     result = await session.stream(
-        text(query),
-        params={
-            "org_id": org_id,
-            "major": str(major),
-            "minor": str(minor),
-        },
+        statement,
+        params={"org_id": org_id, "major": str(major), "minor": str(minor), "string_ids": list(host_groups)},
```
samdoran, line 218: `host_groups` is an iterable (a set). You shouldn't need to convert it to a list because it will get expanded by the BindParameter.

Keep these values on multiple lines to make the code review easier. The real diff here is smaller but it's hidden by the formatting change.

src/roadmap/common.py
```diff
@@ -87,55 +89,133 @@ async def query_rbac(
...
+        #     "updated": "2025-01-07T12:59:52.471612+00:00"
+        #    }
+        # ]
+
+        # There is a special case. If None is in host_groups, we must query
+        # for a group that has ungrouped == True.
+        ungrouped_query = "EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE (group_obj->>'ungrouped')::boolean = true)"
+        # This query searches for any group record with an id that matches any
+        # of our eligible host group ids.
+        grouped_query = "EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE group_obj->>'id' IN :string_ids)"
```
samdoran, line 202: It could be easier to follow if the bind parameter name matched the variable name.
```suggestion
        grouped_query = "EXISTS (SELECT 1 FROM jsonb_array_elements(hosts.groups::jsonb) AS group_obj WHERE group_obj->>'id' IN :host_groups)"
```

src/roadmap/common.py
Lifto: yes, ungrouped is in the database.

confirmed with:
```http -b https://gabi-host-inventory-prod.apps.crcp01ue1.o9m8.p1.openshiftapps.com/query Authorization:"Bearer $(oc whoami -t)" query="select * from hbi.hosts where groups != '[]' limit 1"
```

src/roadmap/common.py
Lifto: fixed

src/roadmap/common.py
Lifto: lint won't let me spread params over multiple lines

Lifto: I have to explicitly make list or I get:
```
sqlalchemy.exc.StatementError: (builtins.TypeError) 'set' object is not subscriptable
```

samdoran: Interesting. I suppose that has to do with how SQLAlchemy is using it internally.

samdoran: > lint won't let me spread params over multiple lines
If you put a trailing comma on the last item, `ruff` will [format it](https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma) so that each item is on its own line.

src/roadmap/common.py
samdoran: I see what this is doing now. Maybe that's idiomatic SQL, but it still seems too clever and hard to follow.

IMO, it's easier to grok the final query statement if CTEs are used to gather the data we want and then query against it. That would also remove the need to inspect the query string to add the bind parameter. And if we switch to an array instead of a list of strings, we don't need an expandable bind parameter at all.

src/roadmap/common.py
Lifto: fixed

src/roadmap/common.py
```diff
@@ -87,55 +89,133 @@ async def query_rbac(
...
+    return allowed_group_ids
 
 
-# FIXME: This should be cached
 async def query_host_inventory(
     org_id: t.Annotated[str, Depends(decode_header)],
     session: t.Annotated[AsyncSession, Depends(get_db)],
     settings: t.Annotated[Settings, Depends(Settings.create)],
-    permissions: t.Annotated[list[dict], Depends(check_inventory_access)],
+    host_groups: t.Annotated[set[UUID], Depends(get_allowed_host_groups)],
```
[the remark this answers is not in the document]
Lifto: fixed

src/roadmap/common.py
Lifto: my case for this style:

* the EXISTS (SELECT 1... style should cease searching once a match has been found. I only tested this on one client I could find, but there is a chance the speedup is greater on clients with tons of hosts and groups.  I've tried to make this efficient code understandable with indentation and comments. (the stats above show it is doing in 1.5ms what CTE style is doing in 3ms, for the one org I'm testing on.)

* the current method of chopping up/assembling the query with Python lets us have no group clauses in queries that do not use groups (probably the majority of queries, but I've not gathered any real stats). I've tried to comment this up.

* It is tricky because in certain cases we have to query for EITHER a group-id match OR a group with ungrouped = true. I think the CTE approach, if it were to be maximally readable, would require us to query for those conditions on every query not just ones where a None or group id were present, and that would be slower. (hence it's a tradeoff.)

I'm asking myself if I came back to this code with no memory and had to understand it, and I had these comments and structure, if I'd be OK with it, and I would be.

beav: "exists (select 1" works for me

samdoran: My overall concern with this code is legibility and maintenance, especially since it is related to access control. I don't love the solutions I have come up with. I would like to arrive at the simplest and most obvious solution, one that we will be able to modify and maintain with confidence several months from now.

samdoran: I would like to rewrite this to use the ORM because it would be easier _for me_ to read and maintain. But I'm not confident that's the best _technical_ solution.

Let's polish what we have and get comfortable with it and be confident it is doing the right thing.

src/roadmap/common.py
```diff
@@ -87,54 +89,183 @@ async def query_rbac(
...
+                # It is possible that the group id is None, which means
+                # permission to access a host who belongs to a group which
+                # has its "ungrouped" attribute set to True.
+                # TODO: Check if this None check can be removed once the Kessel Phase 0 migration is complete.
+                # https://issues.redhat.com/browse/RHINENG-16842
+                if gid is not None:
+                    UUID(gid)
+        except (ValueError, TypeError):
+            logger.warning(f"RBAC attributeFilter contained erroneous UUID: '{gid}'")
+            raise HTTPException(501, detail="Received invalid UUIDs for attributeFilter.value in RBAC response.")
```
samdoran, line 121: If possible, narrow the scope of this try except.
```suggestion
        for gid in group_list:
            # It is possible that the group id is None, which means
            # permission to access a host who belongs to a group which
            # has its "ungrouped" attribute set to True.
            # TODO: Check if this None check can be removed once the Kessel Phase 0 migration is complete.
            # https://issues.redhat.com/browse/RHINENG-16842
            if gid is not None:
                try:
                    UUID(gid)
                except (ValueError, TypeError):
                    logger.warning(f"RBAC attributeFilter contained erroneous UUID: '{gid}'")
                    raise HTTPException(501, detail="Received invalid UUIDs for attributeFilter.value in RBAC response.")
```

src/roadmap/common.py
```diff
@@ -87,54 +89,183 @@ async def query_rbac(
...
+    Only return data the authenticated user is permitted to read. Specifically,
+    if "host_groups" is not empty (which implies unrestricted access), only
+    return hosts that belong to a group present in "host_groups"
+
+    Note there is a special case in the result "get_allowed_host_groups"
+    returns, in that the return value may contain (with or without other group
+    ids) a None value. If None is present in "host_groups", one of the
+    permitted  groups is the "ungrouped" group. While other groups are
+    identified by the value of their "id" field, the "ungrouped" group is
+    identified by the fact that the value of its "ungrouped" field is `true`.
```
samdoran, line 183: Thank you for this doc string.

src/roadmap/common.py
```diff
@@ -87,54 +89,183 @@ async def query_rbac(
...
+        # then the clauses that detect them must be comibined with on OR
+        # statement.
+        if None in host_groups:
+            if len(host_groups) > 1:
+                # Accept either a group id match or ungrouped = true.
+                query = f"{query} AND ({ungrouped_query} OR {grouped_query})"
+            else:
+                query = f"{query} AND {ungrouped_query}"
+        else:
+            query = f"{query} AND {grouped_query}"
```
samdoran, line 260: I find this section to be the most troublesome to reason about. This doesn't lower the overall complexity but I think it makes it a bit easier to read.
```suggestion
        query = f"{query} AND {grouped_query}"
        if None in host_groups:
            query = f"{query} AND {ungrouped_query}"
            if len(host_groups) > 1:
                # Accept either a group id match or ungrouped = true.
                query = f"{query} OR {grouped_query})"

```

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -220,6 +219,77 @@ async def query_rbac_override():
     assert result.status_code == 200
 
 
+def test_get_relevant_app_stream_resource_definitions_with_ungrouped_permission(api_prefix, client):
+    """Testing a case with group None, which means 'ungrouped'"""
```
samdoran, line 223: This is an excellent example of when doc strings are helpful for tests. What you have is fine but another way to write this is with the template "Given [scenario] test/ensure/assert that [expected result]".

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -220,6 +219,77 @@ async def query_rbac_override():
...
+    client.app.dependency_overrides = {}
+    client.app.dependency_overrides[query_rbac] = query_rbac_override
+    client.app.dependency_overrides[decode_header] = decode_header_override
+    result = client.get(f"{api_prefix}/relevant/lifecycle/app-streams?related=true")
+    data = result.json().get("data", "")
+    assert result.status_code == 200
+    assert len(data) == 2
+    # In the test data there is an eligible system from another group (for
+    # which the request does not have permission) that shows NGINX 1.14.
+    display_names = {d["display_name"] for d in data}
```
samdoran, line 288: nit: Put the data gathering above the assert statements for greater legibility.

samdoran:
There are a small number of things to address then we can merge this. Great work!

src/roadmap/common.py
Lifto: fixed

tests/v1/lifecycle/test_app_streams.py
Lifto: fixed

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -220,6 +219,77 @@ async def query_rbac_override():
...
+    assert "Node.js 18" in display_names
+    assert "NGINX 1.22" in display_names
```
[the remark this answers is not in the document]
Lifto: changed it to this, and removed the check for Len == 2
```    assert {"Node.js 18", "NGINX 1.22"} == display_names
```

src/roadmap/common.py
```diff
@@ -87,54 +89,185 @@ async def query_rbac(
...
 
-    return host_permissions
+    for perm in host_permissions:
+        # Get the list of allowed Group IDs from the attribute filter.
+        for resourceDefinition in perm["resourceDefinitions"]:
+            if not resourceDefinition:
+                # Any record with an empty resourceDefinition means
+                # unrestricted access.
+                # https://insights-rbac.readthedocs.io/en/latest/management.html#resource-definitions
+                return set()
```
samdoran, line 157: I don't think there's a way this code can be reached. If `resourceDefinition` is `[]`, then the loop is skipped and this falls through to returning the default value for `allowed_group_ids`, which is an empty set.

I think it's safe to remove this code, or rewrite it a bit if we want to retain the explicit `return`.
```python
    for perm in host_permissions:
        if not (resourceDefinition := perm["resourceDefinitions"]):
            # Any record with an empty resourceDefinition means
            # unrestricted access.
            # https://insights-rbac.readthedocs.io/en/latest/management.html#resource-definitions
            return set()

        # Get the list of allowed Group IDs from the attribute filter.
        for resourceDefinition in perm["resourceDefinitions"]:
            group_list = _get_group_list_from_resource_definition(resourceDefinition)
            allowed_group_ids.update(group_list)
```

Lifto: fixed

samdoran (approved):

PR #185, opened 2025-06-02
[RSPEED-1555] App streams and RHEL have different near retirement values

samdoran:
For RHEL only, a release is considered near retirement when the end date is within three months of the current date.

For app streams, retirement is considered within six months.

I incorrectly set the value to three months for both RHEL and app streams in #181.

### DIFF
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -62,6 +62,7 @@ class Lifecycle(BaseModel):
             start_date=self.start_date,
             end_date=self.end_date,
             current_date=today,
+            months=3,
         )
 
         return self
@@ -119,7 +120,7 @@ def _calculate_support_status(
     start_date: date | None,
     end_date: date | None,
     current_date: date,
-    months: int = 3,
+    months: int,
 ) -> SupportStatus:
     support_status = SupportStatus.unknown
 
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -25,7 +25,7 @@ class LifecycleType(StrEnum):
 
 class SupportStatus(StrEnum):
     supported = "Supported"
-    near_retirement = "Support ends within {number} months"
+    near_retirement = "Near retirement"
     retired = "Retired"
     not_installed = "Not installed"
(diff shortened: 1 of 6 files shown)


src/roadmap/models.py
```diff
@@ -25,7 +25,7 @@ class LifecycleType(StrEnum):
 
 class SupportStatus(StrEnum):
     supported = "Supported"
-    near_retirement = "Support ends within 3 months"
+    near_retirement = "Support ends within {number} months"
```
samdoran, line 28: This is a bit of a hack, using a format string as an enum.

The problem currently is the number is in the enum value. I would like to remove the number from the enum value and make it simply "Near retirement". This would require UI code changes and it also obscures the value considered when calculating the near retirement status.

If we are ok with API just returning "near retirement" and documenting how that is calculated, I would be happy to change the code.

samdoran: Had some discussion and the UI folks are ok with returning "Near retirement" instead of including the month value in the string. I'll work on that change.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -91,6 +91,7 @@ def update_support_status(self):
             start_date=self.start_date,  # pyright: ignore [reportArgumentType]
             end_date=self.end_date,  # pyright: ignore [reportArgumentType]
             current_date=today,  # pyright: ignore [reportArgumentType]
+            months=6,
```
hosekadam, line 94: I was thinking where this happens for rhel - I see it's default value added into `_calculate_support_status` [here](https://github.com/RedHatInsights/digital-roadmap-backend/commit/fc13ff7d0bed8bf47f61343c6ddd9b9406f7984a#diff-7082acc41212109bf5a24f7197b84d1500a833bd89832a3b2088cedc9f72a452R120)

Maybe I would prefer setting it each time manually to be more explicit where which value is used. But I'm definitely fine with this too -> approving those changes

hosekadam (approved):

hosekadam (approved):

PR #193, opened 2025-06-13
[RSPEED-1526] Update module and package data

samdoran:
Update app stream module and package data to incorporate RHEL 9.6, 10.0, and 10.1 data.

The schema in lifecycle-defs changed to include a new attribute `application_stream_type`, with a set number of values. I created a new string enum to hold those values and added it to the relevant response.

With the addition of those schema changes, how a stream is determined to be rolling changed as well. Previously a combination of ACG and end date being `1111-11-11` meant an app stream was rolling. Now it is explicitly set in the `application_stream_type` field.

I made a few minor adjustments to the display names for some app streams.


hosekadam (dismissed):
Looks good to me - the previous generation of the rolling release is removed in favor of the new filed. Unit tests are changed accordingly. Frontend app works fine.

samdoran:
Looking at the data a bit more this morning, there are some funny things going on. I've identified one bug that explains some of the oddities.

I think I'm also going to split the package and module data into separate files. Having these huge variables in a file is a pain to deal with. This was an experiment and I can say that I am not happy with it in practice.

PR #197, opened 2025-06-18
[RSPEED-1515] Change how packages are processed to reduce response time

samdoran:
Add some caching and move processing of packages outside the hosts loop in order to build the response faster. With these changes, 30K hosts in inventory takes about 6 seconds to respond instead of 30.

```
> psql -c 'SELECT COUNT(*) FROM hbi.hosts;'
 count
-------
 29260
(1 row)

> http --meta localhost:8000/api/roadmap/v1/relevant/lifecycle/app-streams
Elapsed time: 5.986918125s
```


samdoran:
The module results have changed. I need to scrutinize that a bit more.

samdoran:
Turns out I had a caching bug. I fixed that as well as one other bug I found.

samdoran:
This is definitely an improvement:

With ~30,000 hosts in inventory, current `main` takes about 30 seconds:
```
> http --meta localhost:8001/api/roadmap/v1/relevant/lifecycle/app-streams
Elapsed time: 29.881970833s
```

This branch is about 6.6 seconds once I fixed the incorrect package processing, making it a tad slower.
```
> http --meta localhost:8000/api/roadmap/v1/relevant/lifecycle/app-streams
Elapsed time: 6.659168583s
```

PR #200, opened 2025-06-23
[RSPEED-1588] Add EUS and E4S dates for RHEL 8.10 and 9.10

samdoran:
I'm not entirely sure this is correct, but it returns dates instead of Unknown, which allows it to be displayed in the UI.

Officially there does not appear to be EUS or E4S versions of RHEL 8.10, but we definitely have customers with E4S on RHEL 8.10.

### DIFF
--- a/src/roadmap/data/systems.py
+++ b/src/roadmap/data/systems.py
@@ -12,6 +12,7 @@ from roadmap.models import RHELLifecycle
 # FIXME: This needs to be in the database
 
 # https://access.redhat.com/articles/3078
+# https://access.redhat.com/articles/4038291
 OS_LIFECYCLE_DATES = {
     "7": RHELLifecycle(
         major=7,
@@ -116,6 +117,7 @@ OS_LIFECYCLE_DATES = {
         minor=2,
         start_date=date(2020, 4, 28),
         end_date=date(2020, 11, 3),
+        end_date_eus=date(2022, 4, 30),
         end_date_e4s=date(2024, 4, 30),
     ),
     "8.3": RHELLifecycle(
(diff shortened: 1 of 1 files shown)


samdoran:
Had some discussion internally and this seems to be the best solution: display the longest end date for the .10 releases for E4S and EUS.

src/roadmap/data/systems.py
```diff
@@ -150,7 +152,7 @@
         major=8,
         minor=7,
         start_date=date(2022, 11, 9),
-        end_date=date(2023, 5, 16),
+        end_date=date(2023, 5, 31),
```
beav, line 155: just confirming, this change is expected?

src/roadmap/data/systems.py
```diff
@@ -191,7 +195,7 @@
         major=9,
         minor=1,
         start_date=date(2022, 11, 15),
-        end_date=date(2023, 5, 10),
+        end_date=date(2023, 5, 31),
```
beav, line 198: same here, this is expected?

src/roadmap/data/systems.py
samdoran: I came across those in the [Retired Life Cycle Dates](https://access.redhat.com/articles/4038291#red-hat-enterprise-linux-8-7) for RHEL 8.

src/roadmap/data/systems.py
samdoran: Same thing here but for [RHEL 9](https://access.redhat.com/articles/4038291#red-hat-enterprise-linux-9-8).

samdoran: When I put in the end dates, if I wasn't sure I just made it line up with the next dot release.

PR #206, opened 2025-06-27
[RSPEED-1515] Change how package data are grouped

samdoran:
The goal of using a set initially was to keep the number package items that needed to be processed to a minimu. But when system_id was added to the mix, it remove any possibility of deduplicating across systems. Furthermore, this seems to have introduced a bug that causes inconsistent matching of packages to hosts.

Change the grouping strategy to collect system IDs into groups based on the package and its major OS version. I think that gets close to the original goal while still helping improve performance.

I changed the schema of the local hosts table used for development and testing to JSONB fields which matches the production hosts database. Once this change merges, run the following to recreate the host table.
```
make stop-db start-db load-host-data
```

### DIFF
--- a/scripts/load_host_data.py
+++ b/scripts/load_host_data.py
@@ -13,12 +13,12 @@ from app_common_python import os
 from faker import Faker
 from sqlalchemy import create_engine
 from sqlalchemy import delete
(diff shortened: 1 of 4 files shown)


pr-watson (approved):
Looks good!

PR #213, opened 2025-07-14
Use `gha-update` to update GitHub actions

samdoran:
Add configuration for `gha-update` and update actions.

The update to the Coverage action may require configuration changes.


samdoran:
There's a waring on the coverage upload step I need to look into.

PR #216, opened 2025-07-18
[RSPEED-1114] Return display_name for relevant systems

hosekadam:
Add the display name field for relevant systems in tailored rhel, appstreams and upcoming.

### DIFF
--- a/src/roadmap/models.py
+++ b/src/roadmap/models.py
@@ -93,15 +93,6 @@ class SystemInfo(BaseModel):
     id: UUID
     display_name: str
 
-    # eq and hash needed for using set() of systems_detail
-    def __eq__(self, other):
-        if not isinstance(other, SystemInfo):
-            return False
-        return self.id == other.id and self.display_name == other.display_name
-
-    def __hash__(self):
-        return hash((self.id, self.display_name))
-
 
 class Lifecycle(BaseModel):
     model_config = ConfigDict(extra="forbid")


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -85,7 +87,8 @@ class RelevantAppStream(BaseModel):
     count: int
     rolling: bool = False
     support_status: SupportStatus = SupportStatus.unknown
-    systems: list[UUID]
+    system_names: list[SystemInfo]
+    systems: list[UUID] = []
```
samdoran, line 91:
```suggestion
    system_names: list[SystemInfo]
    systems: set[UUID] = Field(default_factory=_get_system_uuids)
```

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -101,6 +104,17 @@ def update_support_status(self):
...
+    def populate_systems(self):
+        """
+        Populate systems field using data in system_names.
+
+        Note: this can be removed once the systems field is deprecated.
+        """
+        self.systems = _get_system_uuids(self.system_names)
+
+        return self
+
```
samdoran, line 117: This can be removed in favor of `default_factory`.

src/roadmap/models.py
```diff
@@ -87,6 +95,17 @@ def set_display_name(self):
...
+    def populate_systems(self):
+        """
+        Populate systems field using data in system_names.
+
+        Note: this can be removed once the systems field is deprecated.
+        """
+        self.systems = _get_system_uuids(self.system_names)
+
+        return self
+
```
samdoran, line 108: This can be removed in favor of `default_factory`.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -254,7 +268,7 @@ def related_app_streams(app_streams: t.Iterable[AppStreamKey]) -> set[AppStreamK
 async def systems_by_app_stream(
     org_id: t.Annotated[str, Depends(decode_header)],
     systems: t.Annotated[AsyncResult, Depends(query_host_inventory)],
-) -> dict[AppStreamKey, set[UUID]]:
+) -> dict[AppStreamKey, set[tuple[UUID, str]]]:
```
samdoran, line 271: Rather thane a tuple, make this a set of `SystemInfo`.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -254,7 +268,7 @@ def related_app_streams(app_streams: t.Iterable[AppStreamKey]) -> set[AppStreamK
...
     """Return a mapping of AppStreams to ids of systems using that stream."""
```
samdoran, line 272: This doc string may need updating since the return has changed.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -287,21 +301,21 @@ async def systems_by_app_stream(
         if not installed_packages:
             missing["installed_packages"] += 1
 
-        # Store package name, os_major, and system ID for later processing outside the loop.
+        # Store package name, os_major, system ID and display name for later processing outside the loop.
         # This substantially reduces the time it takes for this function to return.
-        system_id = system["id"]
+        system_key = (system["id"], system["display_name"])
```
samdoran, line 306: This should be a `SystemInfo` object and the variable name should change to match. This isn't a key field.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -459,6 +473,8 @@ async def get_relevant_app_streams(
         if app_stream.app_stream_entity.rolling:
             continue
 
+        system_names = [SystemInfo(id=id, display_name=display_name) for id, display_name in systems]
```
samdoran, line 476: If these were already `SystemInfo` objects, this line could be removed.

src/roadmap/v1/lifecycle/rhel.py
```diff
@@ -171,7 +172,7 @@ async def get_relevant_systems(  # noqa: C901
         lifecycle_type = get_lifecycle_type(installed_products)
 
         # Collect system IDs by major version, minor version, and lifecycle type so we can return those in the response
-        system_id = result["id"]
+        system_id = (result["id"], result["display_name"])
```
samdoran, line 175: Same as above: make this an object.
```suggestion
        system_id = SystemInfo(id=result["id"], display_name=result["display_name"])
```

samdoran (changes requested):

src/roadmap/models.py
```diff
@@ -78,7 +85,8 @@ class System(Lifecycle):
     count: int = 0
     lifecycle_type: LifecycleType
     related: bool = False
-    systems: set[UUID] = set()
+    system_names: list[SystemInfo]
+    systems: list[UUID] = list()
```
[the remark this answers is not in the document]
hosekadam: @samdoran About the `set()` I left a note in the commit message [https://github.com/RedHatInsights/digital-roadmap-backend/pull/216/commits/7ae4a45c11356204a19b9bf326b88867d70a8fe2](there). I was also thinking what is better and those were the reasons why I choosed `list` - in the code were already both variants for systems field.

But no problem changing it to set(), I see the benefit in not having duplicates in it. And thanks for the review! :)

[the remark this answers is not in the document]
samdoran: Yup, I saw that. Thanks for writing such great commit messages!

The issue with duplicates is more about our processing code introducing duplicates, not the database having duplicates (which is not possible). I fixed a bug around this in https://github.com/RedHatInsights/digital-roadmap-backend/pull/201.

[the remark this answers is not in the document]
hosekadam: @samdoran _(first half solved, see the the line "^ okay, I solved")_ 
I don't know how to get the `default_factory` working. Your suggestion doesn't work as there is name error `NameError: name '_get_system_uuids' is not defined` (tried also adding `()` and `(data)`) with the `_get_system_uuids` defined in the System class itself and in the `models.py` to avoid scope issues.

I was able to use lambda function:
```
system_names: set[SystemInfo]
systems: set[UUID] = Field(default_factory=lambda data: _get_system_uuids(data))
```
and
```
system_names: set[SystemInfo]
systems: set[UUID] = Field(default_factory=lambda data: set(system.id for system in data["system_names"]))
```
but is says for both cases (the one bellow is from the function call, but the result is the same for the lambda too):
```
return set(system.id for system in data["system_names"])
                                       ~~~~^^^^^^^^^^^^^^^^
KeyError: 'system_names'
```
I think this should work as in the docs is written "The data argument will only contain the already validated data, based on the [order of model fields](https://docs.pydantic.dev/latest/concepts/models/#field-ordering)" and the `systems` field is after `system_names` (as in the 3rd example of the default_factory).

I added a debugger to the `_get_system_uuids` and found that for some reason the `system_names` aren't present there at the moment of the `default_facotry` call:
```
(Pdb) data
{'name': 'RHEL', 'start_date': 'Unknown', 'end_date': 'Unknown', 'support_status': <SupportStatus.unknown: 'Unknown'>, 'display_name': '', 'major': 1, 'minor': 0, 'count': 1, 'lifecycle_type': <LifecycleType.mainline: 'mainline'>, 'related': False}
```
  
  
^ okay, I solved this by using list. But when I change the list to set, getting the ids doesn't work. The data structure looks correct 
```
{'name': 'RHEL', 'start_date': 'Unknown', 'end_date': 'Unknown', 'support_status': <SupportStatus.unknown: 'Unknown'>, 'display_name': '', 'major': 1, 'minor': 0, 'count': 1, 'lifecycle_type': <LifecycleType.mainline: 'mainline'>, 'related': False, 'system_names': {SystemInfo(id=UUID('03499b8c-d56c-4581-b259-cbdebf19ae92'), display_name='db-75.williams-hall.org')}}
``` 
but I'm getting `TypeError: 'set' object is not subscriptable` from `return set(system.id for system in data['system_names'])` statement. (the similar statement used in `validator` works)

This drives me to a question, what would be a benefit of using `default_factory` instead of the `model_validator` in this case? For me, the model validator is easier to read and more error tolerant - in case anybody switches the order of items later, it might be hard to debug if the person won't know about the order needed for default_factory. Also this weird behavior  (this opinion might be based on the experience I have now with it as it doesn't work - maybe I will fall in love with that later :D)

src/roadmap/v1/lifecycle/app_streams.py
hosekadam: Thanks! Creating the `SystemInfo` right in this place is a great idea!

src/roadmap/models.py
```diff
@@ -11,6 +11,20 @@
...
+    Populate systems field using data in systems_detail.
+
+    Note: this can be removed once the systems field is deprecated.
+    """
+
+    if "systems_detail" in data:  # System and RelevantAppStream
+        return set(system.id for system in data["systems_detail"])
+    elif "potentiallyAffectedSystemsDetail" in data:  # UpcomingOutputDetails
+        return set(system.id for system in data["potentiallyAffectedSystemsDetail"])
+    return set()
```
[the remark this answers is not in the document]
hosekadam: I agree the one-liner is less complex, but for the readability, I would prefer the option without the nested `get` calls. What do you think about:
```
systems_data = data.get("systems_detail") or data.get("potentiallyAffectedSystemsDetail")
if systems_data:
    return {system.id for system in systems_data}
return set()
```

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -554,3 +557,29 @@ def test_from_string(package, expected):
...
+
+
+def test_relevant_app_stream_populate_systems_from_systems_detail():
+    """Check if the systems are correcly set using validator."""
+    system1_id = uuid4()
+    system2_id = uuid4()
+    systems_detail = {
+        SystemInfo(id=system1_id, display_name="System 1"),
+        SystemInfo(id=system2_id, display_name="System 2"),
+    }
```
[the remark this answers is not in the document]
samdoran: To have a fixture that can return data, the fixture returns a callable instead of data.
```python
@pytest.fixture
def make_systems():
    def _make_systems(count=3):
        return [SystemInfo(id=uuid4(), display_name=f"System {n}") for n in range(1, count + 1)]

    return _make_systems
```

Then in your test
```python
def some_test(make_systems):
    systems_detail = make_systems(5)
    ...
```

src/roadmap/models.py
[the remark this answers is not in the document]
samdoran: I don't have a good technical reason why your solution won't work — I think it's a matter of style and preference. I don't think changing the default to an empty list adds much value. The walrus operator might also be useful here.

tests/test_models.py
```diff
@@ -58,3 +63,39 @@ def test_calculate_support_status_system(mocker, current_date, system_start, sys
...
+
+@pytest.mark.parametrize("count", (0, 1, 2))
+def test_system_populate_systems_from_systems_details(generate_system_detail, count):
+    system_ids = set()
+    systems_detail = set()
+
+    for _ in range(0, count):
+        system, system_id = generate_system_detail
+        system_ids.add(system_id)
+        systems_detail.add(system)
```
samdoran, line 76: This look can be eliminated by change the fixture to accept a count parameter.
```suggestion
    systems_detail = generate_system_detail(count)
```

tests/test_models.py
```diff
@@ -58,3 +63,39 @@ def test_calculate_support_status_system(mocker, current_date, system_start, sys
...
+        count=count,
+        start_date=date(2022, 5, 17),
+        end_date=date(2032, 5, 31),
+        systems_detail=systems_detail,
+    )
+
+    assert system_lifecycle.systems == system_ids
+
+
+def test_get_system_uuids():
```
samdoran, line 92: What is the purpose of this test?

hosekadam: Will improve the name. It's testing case when any of the attributes is available (`data.get("systems_detail", []) or data.get("potentiallyAffectedSystemsDetail", [])`) - mostly to get good coverage.

tests/v1/lifecycle/test_app_streams.py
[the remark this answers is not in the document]
hosekadam: That's impressive, thanks! :)

src/roadmap/models.py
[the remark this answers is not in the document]
hosekadam: Removed the explicit default value, I added it there for consistency, but It's not needed. I didn't know about the walrus operator, looks great.

I prefer simpler lines of code which each one is easily readable - I think it better match the clean code practices. But it depends on the experience and how much your eyes are trained :)

tests/test_models.py
hosekadam: @samdoran changed

tests/conftest.py
```diff
@@ -86,3 +88,20 @@ def ids_by_os(read_json_fixture):
...
+def make_systems():
+    def _make_systems(count=3):
+        system_ids = set()
+        systems_detail = set()
+
+        for n in range(0, count):
+            system_id = uuid4()
+            system = SystemInfo(id=system_id, display_name=f"System {n}")
+            system_ids.add(system_id)
+            systems_detail.add(system)
```
samdoran, line 103: This could also be written using comprehensions.
```suggestion
        systems_detail = {SystemInfo(id=system_id, display_name=f"System {n}") for n in range(count)}
        system_ids = {system.id for system in systems_detail}
```

tests/test_models.py
```diff
@@ -58,3 +60,29 @@ def test_calculate_support_status_system(mocker, current_date, system_start, sys
 )
 def test_get_rhel_display_name(name, major, minor, expected):
     assert _get_rhel_display_name(name, major, minor) == expected
+
+
+@pytest.mark.parametrize("count", (0, 1, 2))
```
samdoran, line 65: Does this test still benefit from three different test cases?

tests/v1/lifecycle/test_app_streams.py
```diff
@@ -554,3 +555,23 @@ def test_from_string(package, expected):
         package.release,
         package.arch,
     ) == expected
+
+
+def test_relevant_app_stream_populate_systems_from_systems_detail(make_systems, count=2):
+    """Check if the systems are correcly set using generator."""
```
samdoran, line 561:
```suggestion
    """Check that the systems attribute is set properly by field validation."""
```

src/roadmap/models.py
```diff
@@ -41,6 +41,22 @@ class HostCount(BaseModel):
...
+    display_name: str
+
+    # eq and hash needed for using set() of systems_detail
+    def __eq__(self, other):
+        if not isinstance(other, SystemInfo):
+            return False
+        return self.id == other.id and self.display_name == other.display_name
+
+    def __hash__(self):
+        return hash((self.id, self.display_name))
```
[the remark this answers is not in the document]
samdoran: This is the only remaining blocker.

PR #240, opened 2025-08-27
[RSPEED-1139] Improve detection of installed dnf modules

hosekadam:
Use new status field provided by insights inventory team to get more proper information if the dnf module is actually installed and used.

On RHEL 8 systems are listed all available modules. This results in missing tailoring of the data.

TODO:
- fix unit tests


src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -372,6 +372,11 @@ def app_streams_from_modules(
         if os_major not in get_module_os_major_versions(module_name):
             continue
 
+        if "installed" not in dnf_module.get("status", ["installed"]):
```
samdoran, line 375: We actually want to change this behavior to _exclude_ items where status is missing, but only for RHEL 8. For RHEL 9 and later, only enabled modules will show in the list of `dnf_modules` so we can use them for matching.

We will also need tests for these scenarios to make sure the application is doing the right thing for RHEL 8 with/without status as well as RHEL 9 and later with/without status.

I believe we have hosts like that in the data, but we need to write tests to make sure we don't break the behavior in the future.

hosekadam: > We actually want to change this behavior to exclude items where status is missing, but only for RHEL 8. For RHEL 9 and later, only enabled modules will show in the list of dnf_modules so we can use them for matching.

I'm not sure about this. I don't know when the `status` could be empty array - when I registered system to the insights without any enabled/installed dnf modules, all of them had `status:[default]`. So if we would check for missing status, I think the behavior would be the same as without having the `status` check at all.

And if you mean the default value of "installed" I choose that one to keep the original behavior when the customer hasn't updated the insights data for the system and `status` is not present. Do you think we should rather skip that - can be easily done by using `[]` as default.

About the RHEL9+ I need to check, I feel like from my observation, the dnf module is not listed until is installed/enabled. So the system version check might not be needed, but it's not a problem to add it.

Do we care also about `enabled` dnf modules?

I will definitely add unit tests, totally agree.

samdoran: > I'm not sure about this. I don't know when the status could be empty array

It won't be an empty array, but it may be absent altogether. In that case, for RHEL >= 9 we can assume the module is enabled and include it. For RHEL <= 8, we should switch to excluding modules without `status`.

>  Do you think we should rather skip that

Yes, but only from RHEL <= 8.

> About the RHEL9+ I need to check, I feel like from my observation, the dnf module is not listed until is installed/enabled.

That's correct. This is the behavior we want to have with RHEL 8 as well.

For RHEL 8 systems that _haven't_ updated the client or had their data updated, we need to change the behavior to exclude everything. Currently we assume every item in `dnf_modules` means it is enabled, but that is only the case for RHEL >= 9. The net result is a lot of inaccurate  results for customers with RHEL 8 systems.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -372,6 +372,9 @@ def app_streams_from_modules(
         if os_major not in get_module_os_major_versions(module_name):
             continue
 
+        if os_major <= 8 and "installed" not in dnf_module.get("status", []):
+            continue
+
```
samdoran, line 377: This isn't a complete solution since RHEL 9 or later systems that have modules that are not installed will incorrectly show up in the results. The expression after `and` isn't evaluated because `os_major <= 8` is `False`.

samdoran (changes requested):

PR #251, opened 2025-09-12
[RSPEED-1766] Correct stream for java-11-openjdk on RHEL 8

samdoran:
~Match package to app stream based on number of components in stream. Some app streams only have a single digit (major). Compare package version to the available stream version.~

~This is putting another bandage over the flawed package matching.~

The original change was to make the stream to NEVRA matching based on the length of stream components. I decided to correct the data rather than changing package matching because there is much less risk of unexpected results.


samdoran:
The reason this matches correctly on RHEL 9 but not on RHEL 8 is because the stream on RHEL 8 is "11" and on RHEL 9 is "11.0". This fix seems correct through I worry about the potential for incorrect matching in certain scenarios.

samdoran:
The more I look at this, the more this seems like bad data: stream is incorrect for this package on RHEL 8 in the source data.

PR #284, opened 2026-01-14
Add more exception handling for RBAC

pr-watson:
Due to recent RBAC exceptions in prod, this PR adds additional exception handling for JSONDecodeErrors and Exceptions when querying RBAC.


SerCantus (approved):
Looking good, overall makes the app more resilient

PR #302, opened 2026-03-16
Notificator

hosekadam:


src/notificator/__main__.py
```diff
@@ -0,0 +1,59 @@
...
+from notificator.notificator import Notificator
+from notificator.notificator_config import NotificatorSettings
+from roadmap.custom_logging import setup_logging
+
+
+settings = NotificatorSettings.create()
+setup_logging(json_logs=settings.json_logging, log_level=settings.log_level)
+logger = structlog.get_logger(__name__)
+
+ORG_IDS = [int(os.environ.get("ORG_ID", "1234"))]
```
SerCantus, line 19: This is a list but it seems to always contain one item

src/notificator/kafka.py
```diff
@@ -0,0 +1,121 @@
...
+        dev=settings.dev,
+        bootstrap_servers=settings.bootstrap_servers,
+        topic=settings.notifications_topic,
+    )
+    if settings.dev or settings.bootstrap_servers:
+        producer = _build_producer(settings)
+    else:
+        raise KafkaBrokersNotConfigured("No Kafka brokers configured and dev mode is off")
+
+    await _start_producer(producer)
```
SerCantus, line 99: Random comment, I find it weird that the own AIOKafkaProducer class doesn't implement something similar to `_start_producer`, but I am reading https://aiokafka.readthedocs.io/en/stable/api.html#aiokafka.AIOKafkaProducer and your approach seems correct.

src/notificator/notificator.py
```diff
@@ -0,0 +1,221 @@
...
+        """Fetch hosts from HBI to be able to process relevant systems and appstreams.
+
+        The HBI should be fetched only once for both because of the performance - there can be
+        a huge block of systems registered into inventory and fetching that twice would be
+        time consuming.
+        """
+        start_time = time.time()
+        logger.info("Fetching hosts from inventory", org_id=self.org_id)
+
+        async for session in get_db():
```
SerCantus, line 41: Can the DB have two sessions? I find it a bit weird to get a session like this

SerCantus (commented):
Looking really good, only a few small questions

src/notificator/kafka.py
hosekadam: Thanks for your review! The `_start_producer` is basically just a wrapper around [`aiokafkaproducer.start`](https://aiokafka.readthedocs.io/en/stable/_modules/aiokafka/producer/producer.html#AIOKafkaProducer.start) to have multiple attempts to connect. 

I'm rechecking what [vulnerability-engine](https://github.com/RedHatInsights/vulnerability-engine/blob/master/common/mqueue.py#L142) has and the multi-attempt approach is probably redundant as they don't have it :D But I would still prefer to have it, we can tweak the `KAFKA_RETRY_INTERVAL` to have some time in case of short outage (the same we should do with the notification-backend API later, so there is a bit of self-healing magic :D)

src/notificator/__main__.py
```diff
@@ -0,0 +1,59 @@
...
+
+async def main():
+    await lifecycle_notification()
+    await roadmap_notification()
+
+
+async def lifecycle_notification():
+    failed_orgs = []
+
+    async with kafka_producer() as producer:  # TODO handle (re-raise) exceptiton from start_producer
```
pr-watson:
```suggestion
    async with kafka_producer() as producer:  # TODO handle (re-raise) exception from start_producer
```

src/notificator/notificator.py
hosekadam: I don't think, it's meant to be for one session. I did some experimenting now and I would say this solution comes from claude - I wanted to reuse the code from our backend there. 

I asked once more about this:

> async for session in get_db(): pattern works correctly. There's no functional problem with it.
>
>When you return from inside an async for loop, Python calls aclose() on the async generator, which >triggers the async with async_session() as session: cleanup in get_db(). So the session gets properly >closed and the connection is returned to the pool — exactly as intended.
>
>The only downside is readability: async for on a single-yield generator looks like a loop over multiple >items, when it's really just "give me one session with cleanup." async with communicates that intent >more clearly. But it's a style preference, not a correctness issue.

But you know, as we say in Czechia "dvakrát měř, jednou řež" ~ "measure twice once cut" I checked it by the old way on [stackoverflow](https://stackoverflow.com/a/56162461): 

> `[async for](https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement)` ... simply allows sequential iteration over an async source. For example, you can use async for to iterate over lines coming from a TCP stream, messages from a websocket, or database records from an async DB driver.

So I believe it should be fine. Thanks for pointing this out!

src/notificator/__main__.py
hosekadam: The list is more for the future, as there will be more org_ids to iterate over. If needed, it can be set using env var now - this is going to be changed later. I used this for testing the notificator in ephemeral env

src/notificator/notificator.py
```diff
@@ -0,0 +1,221 @@
...
+
+NOTIFY_STATUSES = {SupportStatus.retired, SupportStatus.near_retirement}
+
+
+class Notificator:
+    settings = NotificatorSettings.create()
+
+    def __init__(self, org_id: int):
+        self.org_id = org_id
+
```
pr-watson:
```suggestion

    def __init__(self, org_id: int):
        self.org_id = org_id
        self.settings = NotificatorSettings.create()
```

pr-watson: Better to add the settings instantiation here so it's created at runtime rather than upon importing

hosekadam: Good catch, thanks!

SerCantus (approved):
Ready to ship 🚀

PR #324, opened 2026-04-10
[RSPEED-2637] Ensure EOL packages are not considered related

pr-watson:
### Problem

NGINX 1.22 (and other EOL app streams) were incorrectly appearing as "related" upgrade paths even when:
1. They are past their End of Life (EOL) date
2. No instances of that app stream family are installed on any systems

**Acceptance Criteria:**
- ✅ NGINX 1.22 is not considered as "related" when it's past EOL
- ✅ Related streams are only shown if `end_date > today` or `end_date is None`

### Root Cause

The `related_app_streams()` function had **inconsistent EOL filtering** between two code paths:

**Case 1** (same RHEL version - newer stream):
```python
if app.end_date is None or app.end_date > date.today():  #  EOL check
    add = True
```
**Case 2** (newer RHEL version - upgrade path):
```python
if app.start_date > app_stream_key.app_stream_entity.start_date: # No EOL check
    add = True  # 
```
Case 1 correctly filtered out EOL streams, but Case 2 did not. This meant EOL streams on newer RHEL versions (like NGINX 1.22 on RHEL 9) appeared as related even when they shouldn't.

Now both cases consistently filter out streams where end_date <= today.

Before:
Screenshot From 2026-04-10 11-56-31
After:
Screenshot From 2026-04-10 11-57-17

### DIFF
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -291,6 +291,25 @@ class ModuleStatus(StrEnum):
     installed = auto()
 
 
+def _is_not_eol(app: AppStreamEntity) -> bool:
+    """Check if an app stream is not past its end of life date."""
+    return app.end_date is None or app.end_date > date.today()  # pyright: ignore [reportArgumentType, reportOperatorIssue]
+
+
+def _should_add_same_rhel_version(app: AppStreamEntity, installed: AppStreamEntity) -> bool:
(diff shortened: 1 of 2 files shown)


SerCantus (approved):
Thanks for the refactor! It made easier to work with that part.

To anyone reviewing the PR, the change is subtle, it adds `_is_not_eol(app)` check on newer RHEL versions.

hosekadam (approved):
Nice improvement! Much easier to understand now

PR #325, opened 2026-04-14
Update the data used for local testing of the Roadmap tab

pr-watson:
Quick PR with updates to the Roadmap test data for better testing the new Added to roadmap filter

### DIFF
--- a/src/roadmap/data/upcoming.json
+++ b/src/roadmap/data/upcoming.json
@@ -2,7 +2,10 @@
   {
     "name": "Add Node.js to RHEL7 AppStream THIS IS TEST DATA",
     "type": "addition",
-    "packages": ["nodejs", "npm"],
+    "packages": [
+      "nodejs",
+      "npm"
+    ],
     "release": "7.2",
     "date": "2019-08-01",
     "details": {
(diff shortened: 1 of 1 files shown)


SerCantus:
Is this data really used in stage? When the image is built the data that is used is the one from the data repository:
https://github.com/RedHatInsights/digital-roadmap-backend/blob/main/Containerfile#L69

hosekadam:
You are right, @SerCantus, this data aren't really used. I think it's just an extension of our testing mocked data for easier development.

src/roadmap/data/upcoming.json
```diff
@@ -73,5 +73,230 @@
       "dateAdded": "2025-04-22",
       "lastModified": "2025-04-22"
     }
+  },
+  {
+    "name": "Python 3.12 EOL",
```
hosekadam: maybe I would append a `THIS IS TEST DATA` to all new entries as it is in previous entries.
```suggestion
    "name": "Python 3.12 EOL THIS IS TEST DATA",
```

SerCantus:
@hosekadam Oh okey, the title is a bit misleading. Then I guess this PR is fine

pr-watson:
Yes this is just the data loaded when we do `make load-host-data` for testing locally, sorry about the title!

hosekadam (approved):

PR #326, opened 2026-04-16
[RSPEED-2880] Fix calculation of affected systems

hosekadam:
Calculate unique affected systems. Without this change, when you have 10 appstreams installed on 2 systems and all are pre-retirement/retired, you would end up with 20 affected systems.


SerCantus (approved):
Great fix!!

PR #327, opened 2026-04-20
Disable notificator

hosekadam:
Set notificator to Feb 31 so it never triggers. This allow us to do release without being blocked by notificator which is not yet finished.


SerCantus (approved):
Funny way to disable it

PR #351, opened 2026-05-26
Increase requested memory for notificator

hosekadam:
This is a safe bet for future. Most probably not needed now.


SerCantus (approved):
Change looks ok, we should take a look after the notificator is sent to see the memory that was used and adjust. Lint is failing no idea why.

pr-watson (approved):

PR #352, opened 2026-05-26
Fetch image references concurrently to improve performance

samdoran:
Update image references is concurrent threads to reduce overall execution time.

Since the operation is I/O bound use a thread pool. Use the future object as a key in order to preserve the order of lines in the original file. That way each chunk of the file can be updated concurrently but is then reassambled in the correct order.

`itertools.batched` is ideal for splitting the file into chunks but that is only available in Python >= 3.12. Use a rough equivalent on older Python versions.

This takes execution time from 7.28s to less than 2s.
```
> time ./scripts/update-konflux-refs.py -f .tekton/pipeline-build.yaml 1>/dev/null

________________________________________________________
Executed in    1.74 secs      fish           external
   usr time  228.15 millis    0.24 millis  227.91 millis
   sys time   68.40 millis    1.95 millis   66.45 millis
```

### DIFF
--- a/scripts/update-konflux-refs.py
+++ b/scripts/update-konflux-refs.py
@@ -206,7 +206,7 @@ def array_split(lines: list[str], number: int) -> t.Iterable[tuple[str, ...]]:
     return batched(lines, chunk_size)
 
 
-def process_chunk(data: t.Iterable[str], max_tag_length: int) -> list[str]:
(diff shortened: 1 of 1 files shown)


hosekadam (approved):
Awesome!! From 20 secs to 3 🥇

PR #359, opened 2026-06-08
[RSPEED-3037] Improve module detection

pr-watson:
### Problem
The previous module detection had a critical flaw: it reported modules as "in use" when they were merely **enabled** but not actually **installed**. 

In RHEL, enabling a module stream (`dnf module enable nodejs:18`) doesn't install any packages - it just marks the stream as the default source. Users could enable a stream but never install anything from it, yet our system would incorrectly show those modules as active on their systems.

### Solution
Implement **two-phase module detection** with package verification:

#### Phase 1: Module Processing
- When a module is marked as `enabled` (but NOT `installed`), check if we have package mapping data for it
- If we have the mapping, store the module and its expected packages for later verification
- Don't immediately count it as "in use"

#### Phase 2: Package Verification  
- Cross-reference the system's installed packages against the module's expected packages
- Only count the module as "in use" if at least one of its packages is actually installed on the system

### Changes

**Core detection logic** ([app_streams.py](src/roadmap/v1/lifecycle/app_streams.py)):
- Added `MODULE_PACKAGES` import - maps `(module_name, os_major, stream)` → set of package names
- Enhanced `app_streams_from_modules()` to mark enabled-only modules for verification
- Added verification logic in `systems_by_app_stream()` to check installed packages
- Modified cache to store either `AppStreamKey` or `(AppStreamKey, packages)` tuples

**Supporting infrastructure**:
- Added `download_modulemd.sh` - downloads and strips modulemd files to minimal format
- Added minimal modulemd YAML files (50% smaller, no YAML syntax errors)
- Added `MODULE_PACKAGES` data structure to support package verification
- Added VS Code Python path configuration for better IDE integration

### Impact

✅ **More accurate detection** - eliminates false positives from enabled-but-unused modules  
✅ **Real usage tracking** - only shows modules where packages are actually running  
✅ **Better data quality** - improved reporting for lifecycle management decisions

### Testing

- [x] Verified module detection works for installed modules
- [x] Verified enabled-only modules without packages are excluded  
- [x] Verified enabled-only modules with packages are included
- [x] Tested with RHEL 8 and RHEL 9 systems
- [x] Coverage: 101/102 tracked modules have package data

### Files Changed

- `src/roadmap/v1/lifecycle/app_streams.py` - Core detection logic
- `src/roadmap/data/__init__.py` - Added MODULE_PACKAGES import
- `download_modulemd.sh` - New script to download minimal modulemd files
- `modulemd_data/` - Minimal modulemd YAML files for RHEL 8/9

Jira link: https://redhat.atlassian.net/browse/RSPEED-3037


scripts/download_modulemd.sh
```diff
@@ -0,0 +1,166 @@
...
+    case "$version" in
+        8)
+            repo_name="Rocky Linux 8 (Latest)"
+            repo_url="https://download.rockylinux.org/pub/rocky/8/AppStream/x86_64/os/repodata"
+            ;;
+        9)
+            repo_name="CentOS Stream 9"
+            repo_url="https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/repodata"
+            ;;
+        *)
```
SerCantus, line 39: Why did you use these repos?

src/roadmap/data/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from .app_streams import AppStreamEntity
 from .app_streams import AppStreamType
+from .module_packages import MODULE_PACKAGES as MODULE_PACKAGES
```
SerCantus, line 5: This import doesn't seem to be used

scripts/download_modulemd.sh
SerCantus, line 1: Nice bash script, it's been long since I didn't review one.
Maybe it could be moved next to `scripts/generate_module_packages.py`

scripts/download_modulemd.sh
```diff
@@ -0,0 +1,166 @@
+#!/bin/bash
+# Script to download modulemd files from RHEL-compatible sources
+# and strip them to minimal format for generate_module_packages.py
```
SerCantus, line 3: I'd add some usage example

scripts/generate_module_packages.py
```diff
@@ -0,0 +1,269 @@
...
+                    continue
+
+                # Module documents have document: modulemd
+                if doc.get("document") != "modulemd":
+                    continue
+
+                # Skip obsolete modulemd versions
+                if doc.get("version", 0) < 2:
+                    continue
+
```
SerCantus, line 72: Is this some kind of incompatible formatting? I don't get why version has to be `< 2`

scripts/generate_module_packages.py
```diff
@@ -0,0 +1,269 @@
...
+        f.write("AppStream Module to Package Mappings\n")
+        f.write("\n")
+        f.write("This file maps (module_name, os_major, stream) -> set of package names.\n")
+        f.write("Used to determine if an enabled module actually has packages installed.\n")
+        f.write("\n")
+        f.write(
+            f"Auto-generated by scripts/generate_module_packages.py on {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n"
+        )
+        f.write('"""\n\n')
+
```
SerCantus: All this `f.write` together look really ugly, I think it can be updated into a multi line string

```
import textwrap

# If you don't like breaking the visual indentation of your Python script, 
# you can use Python's built-in textwrap.dedent(). This automatically strips
# out the leading whitespace from the string before writing it to the file.

f.write(textwrap.dedent(f'''\
    """
    AppStream Module to Package Mappings
    
    This file maps (module_name, os_major, stream) -> set of package names.
    Used to determine if an enabled module actually has packages installed.
    
    Auto-generated by scripts/generate_module_packages.py on {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}
    """
    
'''))
```

scripts/generate_module_packages.py
```diff
@@ -0,0 +1,269 @@
...
+                # Multi-line format for readability
+                f.write(f'    ("{module_name}", {os_major}, "{stream}"): {{\n')
+                for i, pkg in enumerate(sorted_packages):
+                    if i == len(sorted_packages) - 1:
+                        f.write(f'        "{pkg}"\n')
+                    else:
+                        f.write(f'        "{pkg}",\n')
+                f.write("    },\n")
+
+        f.write("}\n")
```
SerCantus: This looks very difficult. Isn't there some prettify print for the classes? Like `pprint.pformat`

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -431,14 +449,42 @@ def app_streams_from_modules(
...
+                matched_module = APP_STREAM_MODULES_BY_KEY.get(cache_key)
+                if matched_module and matched_module.start_date:
+                    app_stream_key = AppStreamKey(app_stream_entity=matched_module, name=module_name)
+                    # Mark with special tuple for package verification
+                    cache[cache_key] = (app_stream_key, expected_packages)
+                    logger.debug(
+                        f"Module {module_name}:{stream} enabled but not installed, "
+                        f"marked for package verification ({len(expected_packages)} packages)"
+                    )
+                    continue
```
SerCantus, line 474: Modifying the original `cache` looks dangerous.

SerCantus (changes requested):

scripts/generate_module_packages.py
pr-watson: Anything below version 2 is considered the old format. The fields we need (like artifacts.rpms) might be in different locations or formats, version 2+ is the standard format

scripts/generate_module_packages.py
pr-watson: Great suggestion!

scripts/generate_module_packages.py
pr-watson: Added pformat() and used less code

src/roadmap/data/__init__.py
pr-watson: The import in __init__.py is a re-export - it imports MODULE_PACKAGES and makes it available via from roadmap.data import MODULE_PACKAGES.

src/roadmap/v1/lifecycle/app_streams.py
pr-watson: Good catch, creating a new data structure to hold the new cache for pending verification of packages

scripts/download_modulemd.sh
pr-watson: Will do!

scripts/download_modulemd.sh
pr-watson: Found they had the most reliable source of yamlmd files for RHEL 8 and 9

hosekadam: I wonder about this too, can't we use the official redhat ones? Something like https://rhsm-pulp.corp.redhat.com/content/dist/rhel9/9/x86_64/appstream/os/repodata/ ? (Not sure if the url is the right one, just found this in my bookmarks and it looks similar to what you have there) But I hope there is something official from redhat even if the url is wrong.

But I trust you this works, I just feel we should use our data as source of truth.

src/roadmap/data/__init__.py
SerCantus: You are right :sweat:

scripts/download_modulemd.sh
SerCantus: Thanks!

scripts/download_modulemd.sh
SerCantus: Thanks!

scripts/generate_module_packages.py
SerCantus: Now looks super clean!

SerCantus (dismissed):

SerCantus (approved):

PR #366, opened 2026-06-11
[RSPEED-3250] Improve app_stream_from_package with depth matching

SerCantus:
While validating the https://github.com/RedHatInsights/digital-roadmap-backend/pull/359 I realised that some streams were not being reported. First I thought it was due the new commit but it might be a bit older.
RSPEED-3250

Discovery
I observed that Node.js 16 was not being reported on RHEL 9. The package I had installed was nodejs-1:16.20.2-8.el9_4.x86_64, looks like the initial Node.js 16 release was 16.14 ( https://github.com/RedHatInsights/digital-roadmap-backend/blob/main/src/roadmap/data/packages.py#L7379 ) but over time it was updated to 16.20. Our matching was comparing 16.14 vs 16.20 and failing, not reporting Node.js 16.
Fix
With this PR I add some smart depth detector to properly compare the streams


src/roadmap/v1/lifecycle/app_streams.py
DnsZhou:
```suggestion
    # FIXME: This approach to getting the stream from the package NEVRA is still imprecise.
```

DnsZhou:
```suggestion
    #        the code is working as intended.
    #
    #        Depth-based matching (see _stream_version_depth) mitigates the worst case:
    #        streams whose name carries only a major version (e.g. "Node.js 16") now
    #        compare on major alone, so minor-version drift no longer causes false negatives.
    #        Streams with a major.minor name (e.g. "NGINX 1.20") still require an exact
    #        major.minor match.
    #
    #        In order to accurately lookup the app stream from a package NEVRA string, we need to
```

DnsZhou (dismissed):
Verified from my local side, after I added node js 16 to the test dataset, now the backend returns the right data properly with `curl http://127.0.0.1:8000/api/roadmap/v1/relevant/lifecycle/app-streams/ | python3 -m json.tool`, also verified on local front-end. I just proposed some doc updates.

Nice fix @SerCantus !

Image

hosekadam (approved):
looks good!

PR #367, opened 2026-06-17
[RSPEED-3175] Check only enabled modules on app_streams_from_modules

SerCantus:
RSPEED-3175

Update the module detection as the installed status can be stale. For example by doing this workflow:
```
sudo dnf module enable mariadb:11.8

# At this point the module is enabled but without packages, is not being reported

sudo dnf module install mariadb:11.8

# The module is not enabled and installed, it also installed the packages so we report it

sudo dnf remove mariadb

# We only removed the packages, the module is still reported as enabled and installed, it is being displayed

sudo dnf module uninstall mariadb:11.8

# The module is now only enabled, we no longer display it
```

With this update, we always display modules when they have a "positive status" (enabled or installed). Modules with "negative status" (disabled) are not reported by Insights so it is a technical limitation outside scope.

### DIFF
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -453,9 +453,11 @@ def app_streams_from_modules(  # noqa: C901
         # a module will be incorrectly included in the results.
         cache_key = (module_name, os_major, stream)
 
-        # Check if module is enabled but not installed - needs package verification
-        if ModuleStatus.enabled in module_status and ModuleStatus.installed not in module_status:
-            # Module is enabled but not marked as installed by DNF
+        # Verify enabled modules via actual package installation.
+        # DNF's "installed" flag can be stale (e.g. after `dnf remove <pkg>`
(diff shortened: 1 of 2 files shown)


DnsZhou (approved):
lgtm, verified from my local env with additional testing data added to my local initial-db.

Test data added: 4 RHEL 9.8 hosts with mariadb:11.8 module in different states:

1. enabled + installed, with mariadb packages — simulates normal dnf module install
2. enabled + installed, NO mariadb packages — simulates stale state after dnf remove mariadb (packages gone but module status not updated)
3. enabled only, with mariadb packages — simulates manual package install after dnf module enable
4. enabled only, NO mariadb packages — simulates dnf module enable without installing

Expected result: Only hosts 1 and 3 (those with actual packages installed) should appear under MariaDB 11.8. Hosts 2 and 4 should be excluded because package verification finds no mariadb packages present, regardless of what the module status claims.

See screenshot; the actual behaviour matches the expected result.

Image

Thanks for the great work, @SerCantus !

PR #368, opened 2026-06-18
[RHIENG-27888] Return deployedDate field

hosekadam:
Return deployedDate field in response - in UI is needed to have when a roadmap item was deployed to production. It's also used for roadmap notification.


SerCantus (approved):
Great fix

PR #370, opened 2026-06-23
[RHINENG-27448] Add the notifications Admin Roadmap endpoints

SerCantus:
## Summary
[RHINENG-27448](https://redhat.atlassian.net/browse/RHINENG-27448)
The main goal for this PR is to implement the Admin roadmap endpoints. However while working on it I realised most of the code was duplicated between this endpoint and the Life Cycle one so I included a refactor to create a endpoint "fabric".

- Extract duplicated admin endpoint logic (request models, error handling, route
  definitions) into a `build_notification_router` factory in
  `notifications/__init__.py`.
- Reduce each notification type (`lifecycle`, `roadmap`) to a ~9-line module that
  declares its `NotificationKind` and delegates to the factory.
- Delete the old monolithic `notificator.py` and wire the new sub-routers in
  `admin/__init__.py`.

## Route changes (BREAKING CHANGES!)

This endpoints are not exposed, we are the only ones using it, that's why I didn't include any version.

| Before                              | After                                          |
|-------------------------------------|-------------------------------------------------|
| `/admin/notificator/subscribed-orgs` | `/admin/notifications/lifecycle/subscribed-orgs` |
| `/admin/notificator/custom`          | `/admin/notifications/lifecycle/custom`          |
| `/admin/notificator/all`             | `/admin/notifications/lifecycle/all`             |
| _(n/a)_                              | `/admin/notifications/roadmap/subscribed-orgs`   |
| _(n/a)_                              | `/admin/notifications/roadmap/custom`            |
| _(n/a)_                              | `/admin/notifications/roadmap/all`               |

[RHINENG-27448]: https://redhat.atlassian.net/browse/RHINENG-27448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/roadmap/admin/notifications/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+        except KafkaBrokersNotConfigured as exc:
+            logger.error("Kafka brokers not configured")
+            raise HTTPException(status_code=503, detail="Kafka brokers not configured") from exc
+        except RuntimeError as exc:
+            raise HTTPException(status_code=500, detail=str(exc)) from exc
+
+    async def _trigger_background(org_ids: list[int] | None = None):
+        try:
+            await _trigger(org_ids=org_ids)
+        except Exception:
```
hosekadam:
```suggestion
        except Exception as exc:
```

src/roadmap/admin/notifications/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+            logger.exception(
+                f"Admin trigger: Unexpected background {kind.label} notification failure",
+                org_ids=org_ids,
+                total_orgs=len(org_ids or []),
```
hosekadam, line 100:
```suggestion
                total_orgs=len(org_ids or []),
                error=str(exc),
                error_type=type(exc).__name__,
```
I would improve the error message a bit there. Might be reduntant, as most probably it gets logged in `_trigger`, but there could be something else

SerCantus: Sure, it can't hurt

hosekadam (approved):
You were fast with applying the suggestion! Looks good, nice refactor 🎉

PR #379, opened 2026-07-06
[RHINENG-28543] Module package match improvement

pr-watson:
**Problem**
Modules that share packages were causing false positive detections. For example, Scala was incorrectly detected as "enabled" when only Maven was installed, because both modules include the jansi package.

Shared packages between scala and maven:

- jansi
- jansi-native
- hawtjni
- hawtjni-runtime

When a system had Maven installed (with jansi) and Scala enabled (but not actually installed), the verification logic would see jansi and incorrectly conclude that Scala was in use.

**Solution**
Updated the package verification logic in app_streams.py to require the primary package (package with the same name as the module) when it exists:

If module has a primary package (e.g., scala for scala module, maven for maven module)

→ Only detect the module if that primary package is installed

If no primary package exists (e.g., container-tools)

→ Use the existing logic (detect if any package from the module is installed)

This approach:

✅ Eliminates false positives for modules with shared packages
✅ Preserves correct detection for modules without primary packages
✅ Maintains backward compatibility
Analysis
Out of 95 RHEL 8 modules in our dataset:

87 modules (92%) have primary packages → will use strict verification
8 modules (8%) have no primary package → will use permissive verification
The 8 modules without primary packages:

389-ds, container-tools (all streams), idm (all streams), javapackages-runtime, pki-deps, python27, rhn-tools, satellite-5-client, virt

Jira: [RHINENG-28543](https://issues.redhat.com/browse/RHINENG-28543)

[RHINENG-28543]: https://redhat.atlassian.net/browse/RHINENG-28543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    after_n_builds: 2
+    after_n_builds: 3
     wait_for_ci: false
 
   require_ci_to_pass: false
--- a/src/roadmap/v1/lifecycle/app_streams.py
+++ b/src/roadmap/v1/lifecycle/app_streams.py
@@ -389,10 +389,17 @@ async def systems_by_app_stream(
             systems_by_stream[app_stream].add(system_info)
 
         # Verify enabled-only modules marked for package verification (only for THIS system)
-        for app_stream_key, expected_packages in modules_pending_verification.values():
-            # Check if this module's expected packages are installed on this system
-            if expected_packages & installed_package_names:
-                # At least one package from the module is installed
+        for cache_key, (app_stream_key, expected_packages) in modules_pending_verification.items():
+            module_name = cache_key[0]
+
+            # To reduce false positives from shared packages (e.g., jansi in both scala and maven),
+            # require the primary package (same name as module) to be installed if it exists
+            if module_name in expected_packages:
+                # Module has a primary package - only count if it's installed
+                if module_name in installed_package_names:
+                    systems_by_stream[app_stream_key].add(system_info)
+            elif expected_packages & installed_package_names:
+                # No primary package exists for this module, check for any package match
                 systems_by_stream[app_stream_key].add(system_info)
                 logger.debug(
                     f"Verified module {app_stream_key.name} on system {system_info.display_name}: "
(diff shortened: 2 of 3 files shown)


.codecov.yml
```diff
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    after_n_builds: 2
+    after_n_builds: 3
```
hosekadam, line 3: I think this is not related to this PR. I understand the reason for this (we are having 3 python versions to run unit tests, but reporting just 2), but I think the coverage would be the same.

Like I'm not against this change, but if later anything goes wrong (I don't expect it there, but rather stay on the safe side), I would mention this in commit message - why this was changed. This can be done pretty easy just by `git commit --amend` and updating the description of last commit.

src/roadmap/v1/lifecycle/app_streams.py
```diff
@@ -343,6 +343,29 @@ def related_app_streams(app_streams: t.Iterable[AppStreamKey]) -> set[AppStreamK
...
+) -> None:
+    """Verify enabled-only modules by checking if expected packages are installed."""
+    for cache_key, (app_stream_key, expected_packages) in modules_pending_verification.items():
+        module_name = cache_key[0]
+
+        # To reduce false positives from shared packages (e.g., jansi in both scala and maven),
+        # require the primary package (same name as module) to be installed if it exists
+        if module_name in expected_packages:
+            if module_name in installed_package_names:
+                systems_by_stream[app_stream_key].add(system_info)
```
hosekadam, line 360:
```suggestion
                systems_by_stream[app_stream_key].add(system_info)
                logger.debug(
                    f"Module {app_stream_key.name} verified via primary package on system {system_info.display_name}"
                )
            else:
                logger.debug(
                    f"Module {app_stream_key.name} not verified, primary package not installed on system {system_info.display_name}"
                )
```
I would add some logging there. We can remove later, but for now might be useful if we need to debug something as this detection is pretty new

hosekadam (approved):

PR #383, opened 2026-07-17
[RHINENG-28540] Use deterministic UUID

hosekadam:
Use deterministic UUID for the same day. In case of failure,
notification generated multiple times on the same day will be sent only
once to customers because of filtering on notification backend side.
    
There is a 24h window where the UUID is tracked on notification side.
This window can be extended if needed.


SerCantus (approved):
Ingenious implementation

PR #384, opened 2026-07-17
[RHINENG-28540] Avoid using cache for results

hosekadam:
Because of large organizations, avoid using cache for storing response from HBI and rather request the data twice.

Results in longer processing time, but much better memory performance for big orgs as it's not needed to save the response to the memory. In past we had exactly those issues and notifications were failing because lack of memory.

This ensures even large orgs won't cause failure. Rather slower, but reliably.


SerCantus (approved):
This is the best effort using the API methods.

Here I paste a few things we can do in the future if this still happens:
- Notificator-specific, count-only paths
Don’t reuse systems_by_app_stream / get_relevant_systems / packages_by_system + get_upcoming_data_with_hosts. Those are built for the API (full SystemInfo sets). The notificator only needs small count payloads. Stream hosts once and accumulate counters only.

- Upcoming: the API menthods use a yield(2000), hoever at the end of the loop we have the full org on memory. Invert the loop, drop packages per host. Instead of building packages_by_system for the whole org:
  - load upcoming items (small)
  - for each host: parse packages → match → bump counts / mark affected → discard packages
  - For notifications you may only need “has any affected system” + type/date.

PR #386, opened 2026-07-22
Add a force_email param to the admin API

SerCantus:
The deterministic UUID PR https://github.com/RedHatInsights/digital-roadmap-backend/pull/383 affects testing as it prevent us from sending multiple emails on a single day. With this PR I add a force_email param that can be set via the admin API and forces the notificator to use a new UUID.


hosekadam (approved):
Perfecto 👌

PR #389, opened 2026-07-27
[RHINENG-29220] Introduce shared packages & use while matching

SerCantus:
RHINENG-29220

## Summary

PHP 8.2 (and any App Stream module whose `MODULE_PACKAGES` entry includes a package literally named the same as the module) stopped appearing on the Lifecycle → App Streams view when installed via the standard `sudo dnf module install php:8.2` command. That command installs the module's default profile (`php-cli`, `php-common`, `php-fpm`, `php-mbstring`, `php-xml`, etc.) but not the bare `php` meta-package, since it isn't part of the default profile.

This is a regression from `e28cc10` (#379), which required a module's "primary" package (same name as the module) to be installed whenever `MODULE_PACKAGES` contains one — even when other module-specific sub-packages were present. That rule was intended to prevent false positives from *genuinely shared* packages (e.g. `jansi`/`hawtjni-runtime`, which appear in both `scala` and `maven`), but it was applied too broadly and penalized any module with a same-named package, regardless of whether that package was actually ambiguous.

## Fix

Replaced the blanket "primary package required" rule with an ambiguity-aware one in `_verify_pending_modules()`:

- Only require the primary/module-name package when **every** matched installed package is ambiguous (i.e. also referenced by at least one *other* module on the same RHEL major version).
- Packages unique to a single module (within that RHEL major version) are trusted as evidence on their own.

Ambiguity is computed **per RHEL major version**, not globally — a package name that only collides with an unrelated module on a *different* major version (e.g. `perl-DBD-MySQL` is its own module on RHEL 8 but also a bundled sub-package of `mysql` on RHEL 9) can never actually collide on a real host, since a single host only ever reports one RHEL major version's modules.

The derived `_shared_package_names()` / `SHARED_PACKAGE_NAMES_BY_OS_MAJOR` helper was placed in `roadmap/data/__init__.py`, matching that module's existing pattern for precomputed derived data structures (`OS_MAJORS_BY_APP_NAME`, `APP_STREAM_MODULES_PACKAGES`, `APP_STREAMS`).

## Changes

- `src/roadmap/data/__init__.py` — added `_shared_package_names()` and `SHARED_PACKAGE_NAMES_BY_OS_MAJOR`
- `src/roadmap/v1/lifecycle/app_streams.py` — rewrote `_verify_pending_modules()` to use ambiguity-aware, os_major-scoped matching
- `tests/v1/lifecycle/app_streams/test_app_streams.py` — added 3 new unit tests:
  - `test_systems_by_app_stream_module_detected_via_unique_subpackages_without_meta_package` — reproduces the reported PHP 8.2 scenario
  - `test_shared_package_names_scoped_per_os_major` — proves cross-os_major collisions aren't falsely flagged as ambiguous
  - `test_shared_package_names_real_data_perl_dbd_mysql_not_ambiguous_on_rhel8` — asserts the real data's `perl-DBD-MySQL`/`mysql` collision isn't ambiguous on RHEL 8

### DIFF
--- a/src/roadmap/data/__init__.py
+++ b/src/roadmap/data/__init__.py
@@ -35,7 +35,26 @@ def _only_app_streams(data) -> set[AppStreamEntity]:
     return app_streams
 
 
+def _shared_package_names(module_packages: dict[tuple[str, int, str], set[str]]) -> dict[int, set[str]]:
+    """Return, per RHEL major version, package names referenced by more than one module.
+
+    These are ambiguous signals (e.g. jansi and hawtjni-runtime appear in both
+    scala and maven) and cannot be trusted on their own to verify a module is
+    in use. Packages unique to a single module are trustworthy evidence.
+    """
+    owners_by_os_major: dict[int, dict[str, set[str]]] = defaultdict(lambda: defaultdict(set))
+    for (module_name, os_major, _stream), pkgs in module_packages.items():
+        for pkg in pkgs:
+            owners_by_os_major[os_major][pkg].add(module_name)
+
+    return {
+        os_major: {pkg for pkg, modules in owners.items() if len(modules) > 1}
+        for os_major, owners in owners_by_os_major.items()
+    }
+
+
 APP_STREAM_MODULES_BY_KEY = {(asm.name, asm.os_major, asm.stream): asm for asm in APP_STREAM_MODULES}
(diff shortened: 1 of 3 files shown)


hosekadam (approved):
I like that!

pr-watson (approved):
Looks great!

PR #397, opened 2026-08-10
[RHINENG-29166] Replace blocking urllib RBAC call with async httpx and bound caches

pr-watson:
- Switch query_rbac() from sync urllib.request.urlopen to async httpx.AsyncClient with a 30-second timeout, preventing the RBAC call from blocking the event loop and adding a 504 on timeout.
- Bound NEVRA.from_string() and app_stream_from_package() caches from unbounded functools.cache to lru_cache(maxsize=100_000) to cap memory growth under sustained load.
- Update all RBAC-related tests to mock httpx instead of urllib.

### DIFF
--- a/src/roadmap/common.py
+++ b/src/roadmap/common.py
@@ -4,12 +4,12 @@ import logging
 import textwrap
 import typing as t
 import urllib.parse
-import urllib.request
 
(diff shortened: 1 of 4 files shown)


SerCantus (approved):
Great improvements 🥇

PR #405, opened 2026-08-14
Properly sort image tags with longer version numbers

samdoran:
With the introduction of version numbers consisting of more than two parts, the current sorting is incorrect. The sequence with the fewest items is ordered first.

For example, two version stings 0.3 (timestamp 1,000) and 0.3.2-0 (timestamp 1,001) have different lengths and the shortest would be sorted first even though version 0.3.2-0 has a newer timestamp.

`(0, 3, 1_000) < (0, 3, 2, 0, 1_0001) —> True`(not what we want)

Comparing the version parts as a fixed length sequence padded with zeros makes the sorting behave as intended.

`(0, 3, 0, 0, 1_000) < (0, 3, 2, 0, 1_0001) —> False`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image tag sorting for versioned and timestamped tags.
  - Supports longer image tags, allowing more complete version identifiers.
  - Provides consistent ordering for tags with varying numbers of version components.
  - Handles nonnumeric tags more reliably during reference updates.
  - Improves the reliability of image reference updates when tags use different version formats or include timestamps.

### DIFF
--- a/scripts/update-konflux-refs.py
+++ b/scripts/update-konflux-refs.py
@@ -113,12 +113,28 @@ def image_tag_sort(value):
         # 0.6-622d10efb15c602b5e47d8fd98e374bb2c45c149
         name_parts = name_parts[:-1]
 
+    # Version tags contain a varying number of parts. Some examples:
+    #   - 0.3.2-0
+    #   - 0.3
+    #   - 0.12.0
+    #   - 0.11.2-0
+    #
+    # Sort using a two item sequence.
+    #
+    # The first item is a tuple of integers padded with zeros to always be
+    # four items long, such as (0, 3, 2, 0). The padding is necessary because
+    # a shorter sequence would stop the comparison process and ignore the
+    # timestamp value.
+    #
+    # The second item is the epoch timestamp.
+    timestamp = value.get("start_ts", 0)
+    number_parts = 4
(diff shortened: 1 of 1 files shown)


scripts/update-konflux-refs.py
```diff
@@ -113,12 +113,25 @@ def image_tag_sort(value):
...
+    # such as (0, 3, 2, 0). The padding is necessary so that all values in
+    # each sequence are compared.
+    timestamp = value.get("start_ts", 0)
+    number_parts = 4
+    pad = [0] * number_parts
     try:
-        version = [int(n) for n in name_parts]
-        version.append(value.get("start_ts", 0))
-        return tuple(version)
+        padded_version = ([int(n) for n in name_parts] + pad)[:number_parts]
```
samdoran, line 133: There's a lot going on in this line:

`([int(n) for n in name_parts] + pad)` --> Generator expression that containes a list comprehension and list concatenation operation.
`[:number_parts]` --> slice operation which causes the generator expression to be evaluated

The result is `padded_version` contains a list, not a generator expression.

hosekadam (approved):
I was a bit worried if the images with `-0` appendix will successfully build. But as I tried, everything works just fine.

Thanks @samdoran for those improvements! It's amazing!

PR #413, opened 2026-08-19
Add Kessel / RBAC v2 workspace-aware host authorization

pr-watson:
Add a Kessel-based authorization path that determines a caller's allowed host groups via the Kessel gRPC Inventory API (enumerating the workspaces a user may inventory_host_view) instead of RBAC v1 /access/?application=inventory.

RBAC v1 remains the default; the Kessel path runs only when settings.kessel_enabled is true (settings flag only, no feature-flag service), mirroring advisor-backend and compliance-backend. On the Kessel path the v1 /access/ call is never made.

Workspace ids map onto the existing host_groups set consumed by query_host_inventory: the org root/default workspace implies unrestricted access (empty set), standard workspaces map to their ids (== groups[].id), and no accessible workspaces is a 403.

- config: add kessel_* settings (enabled, url, insecure, auth, principal domain)
- kessel: new module wrapping kessel-sdk (client, subject, workspace lookups)
- common: split v1 parsing into _allowed_host_groups_v1, add Kessel path, make get_allowed_host_groups the orchestrator dependency
- requirements: add kessel-sdk[auth] and regenerate 3.12/3.13/3.14 locks
- tests: unit tests for the Kessel mapping/subject parsing; migrate endpoint tests from overriding query_rbac to overriding get_allowed_host_groups

## Summary by CodeRabbit

- **New Features**
  - Added Kessel-based authorization for workspace and host-group access.
  - Supports organization-wide workspaces, scoped access, and service-account identities.
  - Added configuration options for Kessel endpoints, TLS, and authentication.
  - Added support for Python 3.12, 3.13, and 3.14 environments.

- **Bug Fixes**
  - Users without accessible workspaces are now denied access consistently.
  - Authorization service failures return a clear gateway error.
  - Existing RBAC v1 authorization remains supported as a fallback.


src/roadmap/common.py
```diff
@@ -165,6 +165,66 @@ async def get_allowed_host_groups(
...
+    try:
+        workspace_ids = set(kessel.host_groups_for(client, subject))
+    except HTTPException:
+        raise
+    except Exception as err:
+        logger.error(f"Error querying Kessel for host groups: {err}", exc_info=True)
+        raise HTTPException(status_code=502, detail="Error communicating with authorization service")
+
+    if not workspace_ids:
+        raise HTTPException(status_code=403, detail="Not authorized to access host inventory")
```
SerCantus, line 208: The `_allowed_host_groups_kessel` function description reads:
> The org's root/default workspace implies org-wide access -> empty set (unrestricted).

But on this part if the set is empty and it returns a 403. So exactly how does it work? Does the "default" workspace have a workspace_id? If a user without workspaces then the `workspace_ids` could be empty, returning a 403.

src/roadmap/common.py
```diff
@@ -165,6 +165,66 @@ async def get_allowed_host_groups(
...
+
+    if workspace_ids & kessel.org_wide_workspace_ids(settings, org_id):
```
SerCantus: Is `if workspace_ids &...` ok? I'd expect `if workspace_ids and...`

src/roadmap/kessel.py
```diff
@@ -0,0 +1,144 @@
...
+    return principal_subject(str(user_id), domain)
+
+
+def host_groups_for(client: t.Any, subject: t.Any) -> list[str]:
+    """Return the workspace ids the subject may view hosts in.
+
+    Enumerates workspaces via the Kessel Inventory ``StreamedListObjects`` API
+    (pagination handled by the SDK) using the ``inventory_host_view`` relation.
+    """
+    return [response.object.resource_id for response in list_workspaces(client, subject, HOST_VIEW_RELATION)]
```
SerCantus: Probably not a really performance improvement but if we are always getting the whole list they recommend to create a list directly instead iterating one-by-one.
https://github.com/project-kessel/kessel-sdk-py/blob/15b4ca9fe653114018139ab97c190c0cdc645a3f/src/kessel/rbac/v2/__init__.py#L287

src/roadmap/common.py
pr-watson: yes, that's correct and intended. A user with zero viewable workspaces should get 403. The unrestricted case is separate — it's triggered by the org-wide workspace being present in workspace_ids, not by it being empty.

from claude:

workspace_ids (line 193) is what Kessel returns — the workspaces the caller is allowed to view. If Kessel returns nothing, the caller can see no hosts → 403 deny (line 200-201).
The return value of the function is a different set — the filter query_host_inventory consumes, where empty means unrestricted. We return an empty set only when the caller can view the org's root/default workspace (an org-wide reader).

src/roadmap/common.py
pr-watson: Yup it's a set intersection, i'll change it  `to not workspace_ids.isdisjoint(kessel.org_wide_workspace_ids(...))`, which reads as an explicit intersection test

SerCantus (approved):
Errors begone :wizard:

